PR #233, opened 2015-10-15
Update comment in butterick.symbols

FakeYou:


suchow:
Good catch, thank you.

PR #264, opened 2016-01-15
Remove bare except clauses

Uran198:
It's not good to have bare except clauses because you can accidentally catch what you didn't intend, for example `KeyboardInterrupt`. This can also hide bugs.


suchow:
Thanks!

PR #267, opened 2016-02-07
Add stub for phrasal adjectives

suchow:


suchow:
Eh, this rule is more involved than it seems at first (compare "hotly contested event" to "curly-haired doctor").

PR #290, opened 2016-02-20
Don't print errors object in json output

marsam:
Also replaced `	print` with `click.echo`, see http://click.pocoo.org/5/quickstart/#echoing


suchow:
Thanks — your contribution is much appreciated.

PR #320, opened 2016-03-07
support variadic file args and stdin

jstewmon:
instead of running the demo when no args are given, use --demo

Adding support for stdin is convenient for checking content on the clipboard. For example: `pbpaste | proselint -`


suchow:
@jstewmon Thank you, much appreciated. Question: you seem knowledgeable about the Click package. Would you be willing to take a stab at #326?

PR #327, opened 2016-03-08
fix click option declarations

### LINKED ISSUE #326: Short flags are broken
Each command (e.g. `--version`) should have a short version with a single dash and the first letter of the command (e.g., `-v`). Right now, the short codes are broken and aren't in the correct format.

Here is the relevant code from `command_line.py`:

```
@click.command()
@click.option('--version/--whatever', default=None)
@click.option('--initialize/--i', default=None)
@click.option('--debug/--d', default=False)
@click.option('--score/--s', default=False)
@click.option('--json/--j', default=False)
@click.option('--time/--t', default=False)
@click.argument('file', default=False)
def proselint(file=None, version=None, initialize=None,
              debug=None, score=None, json=None, time=None):
```

jstewmon:
fixes #326

### DIFF
--- a/proselint/command_line.py
+++ b/proselint/command_line.py
@@ -154,22 +154,17 @@ def show_errors(filename, errors, json=False):
 
 
 @click.command()
-@click.option('--version/--whatever', default=None)
-@click.option('--initialize/--i', default=None)
-@click.option('--debug/--d', default=False)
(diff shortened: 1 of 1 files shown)


michaelpacer:
Excellent, thank you for this.

PR #329, opened 2016-03-08
add short option to show help

### LINKED ISSUE #303: Add -h argument option and double dash single character options
Right now `--help` shows a help message, but `-h` is not recognized. I almost didn't try `--help` after doing `-h`. Also, single character command line options should only have one dash, not two.

jstewmon:
fixes #303


michaelpacer:
lgtm.

PR #339, opened 2016-03-08
Only complain about punctuational exclamation marks. Fixes #322

### LINKED ISSUE #322: trigger happy on exclamation mark warnings
When processing markdown or Rmd files it's quite common to have html comments  like

    

Which presently get counted towards this warning: "More than 30 ppm of exclamations. Keep them under control." Perhaps exclamations within comments should be excluded?

jsenn:
Only count exclamation marks at the end of words in 30ppm check to avoid over-counting. For example, don't count the '!' in an `<!-- HTML comment -->`.

### DIFF
--- a/proselint/checks/leonard/exclamation.py
+++ b/proselint/checks/leonard/exclamation.py
@@ -13,6 +13,8 @@ categories: writing
 Too much yelling.
 
 """
+import re
+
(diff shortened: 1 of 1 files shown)


suchow:
Thanks!

PR #352, opened 2016-03-09
Fix spelling of kaleidoscope

drinks:
This corrects 'kaleidascope' to 'kaleidoscope' in the only location I could find it.


suchow:
Thanks, good catch.

PR #353, opened 2016-03-09
Fix backwards reference

drinks:
This flips the source list and preferred usage for 'All of the other reindeer,' which appears to have been implemented backwards.


suchow:
Good catch, thanks.

Issue #242, opened 2015-10-20
Add rule on "not guilty beyond a reasonable doubt"


suchow:
```
source:     Garner's Modern American Usage
source_url: http://www.amazon.com/Garners-Modern-American-Usage-Garner/dp/0195382757
```
>  not guilty (2).

> Today: * "Not guilty beyond a reasonable doubt." 

> This phrasing is ambiguous. The standard by which a jury decides criminal charges is this: a defendant is guilty only if the evidence shows, beyond a reasonable doubt, that he or she committed the crime. Otherwise, the defendant is not guilty. Thus, we say that a defendant was not found "guilty beyond a reasonable doubt."

> But it doesn't follow that we should also say that a defendant was found *"not guilty beyond a reasonable doubt." Is that "not guilty (beyond a reasonable doubt)" or "not guilty-beyond-a-reasonable-doubt"? The latter idea makes more sense — e.g.: "The question is whether a judge can reach a contrary conclusion on the second charge — deciding that though a defendant was not guilty beyond a reasonable doubt, he nonetheless probably committed the crime." "High Court's Highhanded Decision," Chicago Trib., 26 Jan. 1997, at 20. 

> Yet many readers will misconstrue the phrase. Thus, regardless of the writer's intention, some will think of * "not guilty beyond a reasonable doubt" as a strong vindication — rather than as the slight vindication it is (we, the jury, had the slightest bit of reasonable doubt, so we had to find the defendant not guilty). The writer might have gotten it right in the following sentence, but nonlawyers are likely to be misled: "When you know all the facts [of the O.J. Simpson case], you'll see that the prosecutors failed to meet their burden of proof, and how, contrary to the court of public opinion, the jury arrived at their verdict of 'not guilty beyond a reasonable doubt.' " Patricia A. Jones, "Uncensored: Authors Answer Questions Left with Simpson Verdict," Tulsa World, 1 Dec. 1996, at G5.

> If somebody is found not guilty, say "not guilty." Omit the standard ("beyond a reasonable doubt") to prevent a miscue.

> *Invariably inferior form.

tkmharris:
I wrote the check for this (I just copied a similar check and modified it appropriately). I've never worked on an open source project before. How should I contribute the code?

PR #358, opened 2016-03-09
Added check for 'not guilty beyond a reasonable doubt' as raised in  #242.

tkmharris:
I also made it check for 
'not guilty beyond reasonable doubt'
and
'not guilty beyond any reasonable doubt',
as these are both common and equally ambiguous.


suchow:
Thanks for the contribution — merging this now.

Issue #238, opened 2015-10-17
Exception TypeError: "'NoneType' object is not callable" in  ignored


alphapapa:
Well...oops.  :(

```bash
$ pip install --user proselint
$ proselint --debug /tmp/argh.txt
Deleting the cache...
checks.garner.archaism.check
checks.misc.chatspeak.check
checks.consistency.spacing.check
checks.garner.dates.check_dash_and_from
checks.garner.dates.check_decade_apostrophes_long
checks.garner.dates.check_decade_apostrophes_short
checks.garner.dates.check_month_of_year
checks.garner.dates.check_month_year_comma
checks.pinker.apologizing.check
checks.garner.back_formations.check
checks.wsj.athletes.check
checks.misc.annotations.check
checks.misc.eponymous_adjectives.check
checks.consistency.spelling.check
checks.garner.jargon.check
checks.misc.currency.check
checks.garner.commercialese.check
checks.szekely.nword.check_repeated_exclamations
checks.garner.am_pm.check_lowercase_periods
checks.garner.am_pm.check_midnight_noon
checks.garner.am_pm.check_redundancy
checks.garner.am_pm.check_spacing
checks.lilienfeld.terms_to_avoid.check_lie_detector_test
checks.lilienfeld.terms_to_avoid.check_mental_telepathy
checks.lilienfeld.terms_to_avoid.check_p_equals_zero
checks.inc.corporate_speak.check_repeated_exclamations
checks.leonard.hell.check_repeated_exclamations
checks.garner.illogic.check
checks.garner.illogic.check_coin_a_phrase_from
checks.garner.illogic.check_without_your_collusion
checks.palahniuk.suddenly.check_ellipsis
checks.norris.denizen_labels.check
checks.garner.preferred_forms.check
checks.garner.preferred_forms.check_able_atable
checks.garner.preferred_forms.check_able_ible
checks.garner.preferred_forms.check_em_vs_em_and_en_vs_in
checks.pinker.hedging.check
checks.garner.capitalization.check
checks.garner.capitalization.check_days
checks.garner.capitalization.check_months
checks.garner.airlinese.check
checks.ogilvy.pretension.check_repeated_exclamations
checks.garner.sexism.check
checks.pinker.latin.check
checks.wallace.uncomparables.check
checks.garner.animal_labels.check
checks.garner.denizen_labels.check
checks.twain.damn.check_very_damn
checks.misc.hyperbolic.check
checks.pinker.narcisissm.check
[...]

PR #361, opened 2016-03-09
Fix errors on exit

richard-fairthorne:
Addresses issue #238


slang800:
No test cases? :cry: 

...I think you might be better off just finding out where the error is getting thrown from and fixing it there, rather than swallowing everything with a try/catch.

PR #374, opened 2016-03-10
Avoid over-use of the word "actually" with proselint

joshumax:
Actually, I was tired of actually having to actually see people actually write like this...

### DIFF
--- a/README.md
+++ b/README.md
@@ -130,6 +130,7 @@ You can disable any of the checks by modifying `.proselintrc`.
 | `leonard.exclamation` | Avoiding hyperbolic use of exclamation |
 | `leonard.hell` | Avoiding a common cliché |
 | `lilienfeld.terms_to_avoid` | Avoiding misused psychological terms |
+| `misc.actually` | Avoiding the word actually |
 | `misc.annotations` | Catching annotations left in the text |
 | `misc.chatspeak` | Avoiding lolling and other chatspeak |
 | `misc.credit_card` | Keeping credit card numbers secret |
--- a/proselint/.proselintrc
+++ b/proselint/.proselintrc
@@ -39,6 +39,7 @@
         "leonard.exclamation"           : true,
         "leonard.hell"                  : true,
         "lilienfeld.terms_to_avoid"     : true,
+        "misc.actually"                 : true,
         "misc.annotations"              : true,
         "misc.chatspeak"                : true,
         "misc.credit_card"              : true,
--- a/proselint/checks/misc/actually.py
+++ b/proselint/checks/misc/actually.py
@@ -1,0 +1,38 @@
+# -*- coding: utf-8 -*-
+"""Actually, this checks for the overuse of actually.
+
+---
+layout:     post
+source:     unknown
+source_url: http://bit.ly/1FANcFh
+title:      actually
+date:       2014-06-10 12:31:19
+categories: writing
+---
+
+Actually, this checks for the overuse of actually.
+
+"""
+import re
+
+from proselint.tools import memoize
+
+
+@memoize
+def check(text):
+    """Check the text."""
+    err = "misc.actually"
+    msg = u"Actually, you should be careful about overuse of \"actually.\""
+
+    regex = r"[Aa]ctually"
+
+    count = len(re.findall(regex, text))
+    num_words = len(text.split(" "))
+
+    ppm = (count*1.0 / num_words) * 1e6
+
+    if ppm > 5040:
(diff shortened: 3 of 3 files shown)


suchow:
Thanks for contributing. 

Actually, I think we'll need to restrict the scope of this a bit before including it — I don't want to get in the habit of actually banning words unless there is a good reason to never use them (e.g. `garner.skunked_terms`). On the other hand, if it is actually a common tic, there's good reason to point it out.

What do you think about flagging the word when it appears at least three times in a document? Or, actually, we could set a threshold in terms of parts per million, like we do with exclamation marks actually (see `leonard.exclamation.30ppm`)?

joshumax:
@suchow Alright, that actually sounds good to me, I'll set a decent threshold and push.

joshumax:
@suchow Alright, so after a bit of testing it appears that 5,040 ppm is a good threshold for the linter for warning about the over-use of actually... Thoughts?

michaelpacer:
Actually, could you elaborate more about what testing you did?

If we're going to make a single life hacker article the bar to meet in
terms of authority (no insult intended to the author, it's just one article
is not quite the same as writing exhaustive manuals about writing and
style…) we should at least have some standard method for figuring out what
these sorts of thresholds should look like. I actually fear people may see
this rule as potentially producing false positives, since the word (though
it can be overused) actually *can* carry vital meaning.

If you've figured out a good way to test for this kind of threshold that
doesn't end up overdiagnosing the problem, that *actually* will be a way
more important contribution than this (or any) particular rule.

More theoretically, might there actually be a threshold for the overuse of
any particular word? If so how could that be established? You don't
actually have to answer this part it's just some food for thought.

Then on a slightly different note, I feel like there should be a mechanism
for contributing rules that may not be included in the default set. I feel
mixed about this even as a thresholded version. But I do feel like it
should be available to people if they know to look for it, even if we don't
actually include it in the default set. I don't think we actually have
any easy way to do that right now.

I should mention also that until this paragraph I don't actually think I
overused the word “actually” in this comment. Each case was actually
a slightly different meaning and context and I actually think that matters.
This may actually make a test like this one much closer to requiring some
basic nlp capabilities — if it isn't actually AI complete. What is actually
jumping out to me about this paragraph is actually the clustering rather
than the average rate of use. Not actually sure how that should be defined
though. Maybe an proportion of “actually” to sentence counts rather than
word counts? Not saying that is actually the right answer just that it
actually is a possibility. But this paragraph is actually problematic. So
having something that could actually catch this while actually avoiding
flagging the previous paragraphs would actually be damn impressive.

suchow:
Mike, I think what we're hitting upon with "actually", "very", "really", "obviously", &c., is a set of tics that when used sparingly by an author suggest intentionality, but when used often suggest that it's an unconscious habit. We should be in the business of pointing out the bad habit. As an example of this, I recently ran `proselint` over a popular open-source book on Github and found more than 50 instances each of "obviously" and "very", almost none of which were needed.

Garner and others call these "weasel words", and writes:

> WEASEL WORDS. Theodore Roosevelt said, in a speech in St. Louis on May 31, 1916: “One of our defects as a nation is a tendency to use what have been called weasel words. When a weasel sucks eggs it sucks the meat out of the egg and leaves it an empty shell. If you use a weasel word after another there is nothing left of the other.” Some writers have incorrectly assumed that the METAPHOR suggested itself because of the wriggling, evasive character of the weasel. In any event, sensitive writers are aware of how supposed intensives (e.g., very) actually have the effect of weakening a statement. Many other words merely have the effect of rendering uncertain or hollow the statements in which they appear. Among these are candidly, clearly, compelling, duly, frankly, if practicable, manifestly, meaningful, obviously, perfectly, quite, rather, reasonable, seriously, significantly, somewhat, substantially, undue, and virtually. See clearly, obviously, somewhat & very.
>
> Garner, Bryan (2009-07-28). Garner's Modern American Usage (p. 853). Oxford University Press. Kindle Edition. 

They're incorporated into a number of other linters. 

I think the Proselintian way to do this is to be conservative. These words all have legitimate uses, but there is a certain volume of them that is characteristic of overwrought, musteline prose. What if we created a `weasel_words` module with a list of these words. We can then set some threshold, defined by some combination of ppm and count, which triggers one of three states, in order of how confident we are:

1. Don't flag any.
2. Flag only the first occurrence and warn that *x* more were found later.
3. Flag them all.

michaelpacer:
I support that model and would like to suggest revisiting this. 

I still think that having some idea of what the test set for establishing a baseline is. I think this would be good to start doing in general.

PR #377, opened 2016-03-10
Fix the Vim plugin

### LINKED ISSUE #340: Add tex support for vim plugin
Hi,

First, proselint is a wonderful piece of code. I find it very useful.

However, it lacks the support of tex files:

https://github.com/amperser/proselint/blob/master/plugins/vim/proselint.vim

Here you only support text files, but tex files would be great too (because you often write prose in LaTex).

Changing text to tex in all the fille works, but it would be great to add it to proselint in a proper way.

lcd047:
* Turn `proselint.vim` into an external checker for syntastic
* Add support for filetypes `asciidoc`, `docbk`, `help`, `html`, `markdown`, `nroff`, `pod`, `rst`, `tex`, `texinfo`, and `xhtml`.
* Fix message encoding, add loclist sorting, fix defaults.
* Update `readme`.

Closes #340.

### DIFF
--- a/plugins/vim/readme.md
+++ b/plugins/vim/readme.md
@@ -1,29 +1,41 @@
-# Proselint plug-in for Vim
+# Proselint plugin for Vim
 
-Proselint plug-in for Vim is meant to work with
-[Syntastic](http://vimawesome.com/plugin/syntastic), as well as `proselint`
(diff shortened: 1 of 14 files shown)


suchow:
Thanks!

Issue #378, opened 2016-03-10
Multi-word checks do not break over new lines


tkmharris:
Many checks for multiple-word phrases do not flag if the phrase includes a line break.

For example, orwell.debased flags
'It is a consideration which we should do well to bear in mind in these times.'
but not
'It is a consideration which we should do well to bear 
in mind in these times.'

This could be fixed for each check, but it would probably be better to modify the tools.

PR #379, opened 2016-03-10
Added a check for things like 'wax lyrically' vs 'wax lyrical' as in Fowlers.

tkmharris:
Checks like this would be neater to implement if there were a tool that takes a pair of lists of words that form bad combinations (in this case [wax, waxed,...] and [lyrically, poetically, .....]) and flags when it sees one followed by another. This tool would need to take into account issue #378.


suchow:
@tkmharris, can you split this into three PRs, one for 'not guilty...', one for wax &c., and one for inferior/superior? Also please squash each to one commit. I can help with this if needed.

PR #385, opened 2016-03-12
Add recommendations from Edward P. Bailey

david-a-wheeler:
Add recommendations from the book "Plain English at Work"
  by Edward P. Bailey, Jr., Oxford University Press, 1996.

### DIFF
--- a/proselint/.proselintrc
+++ b/proselint/.proselintrc
@@ -1,6 +1,7 @@
 {
     "max_errors": 1000,
     "checks": {
+        "bailey.preferred_forms"        : true,
         "butterick.symbols"             : true,
         "carlin.filth"                  : true,
         "consistency.spacing"           : true,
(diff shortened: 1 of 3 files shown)


suchow:
Hi David, thanks for the contribution. I think it's unlikely we'll be able to merge this in its current form — I don't want us to get in the habit of banning words. On the other hand, I do see value in using plain language when possible. Maybe there's a way to have this check trigger only if it finds many examples of the non-preferred forms in a single document? Do you know of any good examples that we could use to test it?

david-a-wheeler:
I imagine some sort of frequency should be easy to determine. Is there already routine to do that?

Also

david-a-wheeler:
I suggest modifying the contribution information to state that you don't want to get in the habit of banning words. Obviously you do it, with words like very.

david-a-wheeler:
Well, the "Wondermark" awards are rewards for especially writing: http://centerforplainlanguage.org/wondermark/

One "winner" is here:
https://www.uscis.gov/humanitarian/consideration-deferred-action-childhood-arrivals-process/frequently-asked-questions

I could resubmit with some sort of "% words triggering", using this as an example of something that should trigger it.

david-a-wheeler:
I intend to create a new version, based on these comments, so closing this pull request.

PR #392, opened 2016-03-14
Added check for inferior/superior to/than.

tkmharris:
I hope this is better (sorry, still getting used to git).


suchow:
Perfect, thanks.

PR #398, opened 2016-03-14
Fix the Vim plugin to work with older versions of syntastic

lcd047:
* Syntastic has changed since the last official release 3.7.0.  Make the plugin work with older versions.
* Drop support for filetype `docbk`.  Older versions of syntastic don't have the necessary infrastructure to support `docbk` as an external checker.


suchow:
Thanks!

Issue #98, opened 2015-03-25
Add rule about starting paragraph with "But"


suchow:
```
Justin Jungé to me
Show more
Mar 23
Hi Jordan,

I'm grading more tutorial essays and keep making a similar set of comments about a problem that proselint might be able to flag.  

Here's one instance of the comment:

Don't start a paragraph with "but."

Every paragraph should be self-contained.  "But" implies referring back to the previous paragraph to know what contrast is being made. 

On a similar note, the first sentence of a new paragraph should use nouns instead of pronouns.  Pronouns often require referring back to previous sentences, breaking self-containment.

In this particular case, "but" can simply be deleted, because the sentence is sufficiently informative (effectively self-contained).  I'm merely flagging this use to explain why style guides forbid using "but" at the start of a paragraph.


Flagging paragraphs that start with "but" should be trivial to implement.  Recognizing paragraph first-sentences with only pronouns and no nouns also seems feasible, though it won't catch every version of the general infraction.  
…
```

mixmastamyk:
Hi,
This is giving me false alarms.  I'm checking a restructured text file and proselint thinks that if a line starts with "but" it is the start of a paragraph.  This is not correct, because under rst it must have a blank line in front of it to be a new paragraph.  Rst like markdown normalizes whitespace between blank lines in output.

PR #404, opened 2016-03-15
Add check for 'But' at start of paragraph

dhan12:
Added module to check for 'But' at the start of a paragraph. 
This addresses issue #98 .


suchow:
Looks great, thanks. I made a few minor line-by-line comments. Mind updating the PR?

dhan12:
Updated code (and file names) to match comments. 
I had a few commits, so I tried to clean things up by squash-ing them all into one commit. 
Unfortunately, it seems like this has hidden your previous comments.

suchow:
Thanks!

PR #406, opened 2016-03-18
Set option name to json output

marsam:
This avoids shadowing the value of the option json with the json module

### DIFF
--- a/proselint/command_line.py
+++ b/proselint/command_line.py
@@ -113,9 +113,9 @@ def clear_cache():
         shell=True)
 
 
-def show_errors(filename, errors, json=False, compact=False):
+def show_errors(filename, errors, output_json=False, compact=False):
     """Print the errors, resulting from lint, for filename."""
-    if json:
+    if output_json:
         out = []
         for e in errors:
             out.append({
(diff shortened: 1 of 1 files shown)


proselint/command_line.py
```diff
@@ -113,9 +113,9 @@ def clear_cache():
         shell=True)
 
 
-def show_errors(filename, errors, json=False, compact=False):
+def show_errors(filename, errors, json_output=False, compact=False):
```
suchow: Can you rename this to `output_json`?

proselint/command_line.py
```diff
@@ -158,13 +158,13 @@ def show_errors(filename, errors, json=False, compact=False):
 @click.option('--debug', '-d', is_flag=True)
 @click.option('--clean', '-c', is_flag=True)
 @click.option('--score', '-s', is_flag=True)
-@click.option('--json', '-j', is_flag=True)
+@click.option('--json', '-j', 'json_output', is_flag=True)
```
suchow: What does this change do?

suchow:
Thanks! I left two small line comments with a suggested change and a question.

proselint/command_line.py
marsam: Changes the [parameter name](http://click.pocoo.org/5/parameters/#parameter-names) from the default `json` to `json_output`

marsam:
I've changed the parameter it to `output_json`

suchow:
Great, thank you.

Issue #313, opened 2016-03-07
Could not open cache file


mrzl:
Hey,
I just shortly wanted to check this out and ran into a couple of problems. First, one you're probably already aware of, the incompratibility with python 2.7.
I'm running in these problems here when I'm trying to run it with python 3.4 (I tried pip version, current master and v0.3.6 branch):

`
Could not open cache file /.proselint/checks.misc.chatspeak.check, maybe name collision
`
More here: http://pastebin.com/ARZpLDQQ

Lots of these for all different check files, which I can't seem to find anywhere in the repository. This is then followed by ending up here: https://github.com/amperser/proselint/blob/master/proselint/tools.py#L50

```
Traceback (most recent call last):
  File "/root/.venvs/prose/lib/python3.4/site-packages/proselint-0.3.6-py3.4.egg/proselint/tools.py", line 50, in wrapped
    return cache[key]
TypeError: 'NoneType' object is not subscriptable

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/root/.venvs/prose/bin/proselint", line 9, in <module>
    load_entry_point('proselint==0.3.6', 'console_scripts', 'proselint')()
  File "/root/.venvs/prose/lib/python3.4/site-packages/click/core.py", line 716, in __call__
    return self.main(*args, **kwargs)
  File "/root/.venvs/prose/lib/python3.4/site-packages/click/core.py", line 696, in main
    rv = self.invoke(ctx)
  File "/root/.venvs/prose/lib/python3.4/site-packages/click/core.py", line 889, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/root/.venvs/prose/lib/python3.4/site-packages/click/core.py", line 534, in invoke
    return callback(*args, **kwargs)
  File "/root/.venvs/prose/lib/python3.4/site-packages/proselint-0.3.6-py3.4.egg/proselint/command_line.py", line 165, in proselint
    errors = lint(file, debug=debug)
  File "/root/.venvs/prose/lib/python3.4/site-packages/proselint-0.3.6-py3.4.egg/proselint/command_line.py", line 79, in lint
    result = check(text)
  File "/root/.venvs/prose/lib/python3.4/site-packages/proselint-0.3.6-py3.4.egg/proselint/tools.py", line 59, in wrapped
    '%s; it probably has unhashable args'] % (call_to))
TypeError: unsupported operand type(s) for %: 'list' and 'str'
```

Am I doing something wrong?

[...]

PR #408, opened 2016-03-18
Fixed exception in the TypeError handler when cache sync fails

saul:
Fixed attempting to perform string interpolation on a list:
```
  File "/usr/local/lib/python3.5/site-packages/proselint/tools.py", line 59, in wrapped
    '%s; it probably has unhashable args'] % (call_to))
TypeError: unsupported operand type(s) for %: 'list' and 'str'
```


suchow:
Perfect, thank you. This is related to #313.

PR #412, opened 2016-03-22
added tests for butterick.symbols

joshmgrant:
I've added a first set of new tests for butterick.symbols. I was able to add (basic) tests for all checks except for check_curly_quotes.


suchow:
Thanks!

suchow:
Merged by #413.

PR #433, opened 2016-04-02
Add spaCy NLP as a tool

suchow:
(This is in progress and is not ready to merge.)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,6 +5,7 @@ python:
 - '3.4'
 - '3.5'
 install:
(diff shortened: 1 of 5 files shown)


michaelpacer:
Before building any linting rules based on this, we should come up with a few tests that it needs to pass that possibly hinge on its running speed, since the NLP stuff that this is able to produce could dramatically slow down each pass of proselint.

suchow:
Closing for now, will revisit.

PR #438, opened 2016-04-02
Build on both Linux and OS X

suchow:


suchow:
Nope, no Python on OS X. Will try again later...

PR #439, opened 2016-04-03
Handle missing dbm module on some python 2.7 systems

CraigKelly:
I didn't realize it, but apparently some systems don't have the dbm module for Python 2.7. Added a fallback to anydbm, whose error tuple should still do what we want.

### DIFF
--- a/proselint/tools.py
+++ b/proselint/tools.py
@@ -8,11 +8,15 @@ import sys
 import traceback
 import os
 import shelve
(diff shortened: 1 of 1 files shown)


suchow:
Thanks. (Don't mind the AppVeyor check, I'm in the midst of setting it up.)

suchow:
Merged through #440.

PR #446, opened 2016-04-05
Allow proselintrc to be overridden by user

Tatsh:
Check paths `/etc/proselintrc`, package copy of `.proselintrc`, and `~/.proselintrc` for check settings. The values in `~/.proselintrc`, if it exists, will take precedence over the defaults (merge over).

Also, I wanted to add a configuration option that defaults to `~/.proselintrc` but I was not sure of the best way to do that with the Click API.

### DIFF
--- a/proselint/command_line.py
+++ b/proselint/command_line.py
@@ -44,6 +44,7 @@ def run_initialization():
         except Exception:
             pass
 
+
 def load_options():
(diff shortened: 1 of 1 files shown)


suchow:
Thanks! This is great. I'm closing it only because I've updated your code to reflect some recent refactoring, and opened a new PR with your commit, #451.

PR #447, opened 2016-04-05
Do not open file for reading and writing

Tatsh:
This makes files that are readable but not owned by the user able to be checked (mode `0o644`, etc).


suchow:
Thank you.

PR #467, opened 2016-04-12
Remove `shell=True`

suchow:


proselint/command_line.py
```diff
@@ -37,20 +37,24 @@ def timing_test(corpus="0.1.0"):
     for file in os.listdir(corpus_path):
         filepath = os.path.join(corpus_path, file)
         if ".md" == filepath[-3:]:
-            subprocess.call(
-                "proselint {} >/dev/null".format(filepath), shell=True)
+            subprocess.call(["proselint", filepath, ">/dev/null"])
```
QuLogic: This doesn't really work. Without using `shell=True`, the `>` is just another character, so you're just passing an extra file name. You can redirect using something like `stdout=subprocess.DEVNULL`.

tests/test_demo.py
```diff
@@ -2,4 +2,4 @@
 
 import subprocess
 
-subprocess.call("proselint --debug >/dev/null", shell=True)
+subprocess.call(["proselint", "--debug", ">/dev/null"])
```
QuLogic: The same thing applies here (though you've added `shell=True` back on `master`).

suchow:
@QuLogic Thanks for the review! Would you be interested & willing to open a PR that does all this correctly?

Issue #480, opened 2016-05-03
Checks in misc/capitalization.py not catching preferred forms correctly


joshmgrant:
Using a Win 10 machine, I did the following: 

- Create a TXT test input file called pref_prose_test.txt (included)
- Run the command "proselint pref_prose_test.txt"
- Look at the output

The expected result is that all incorrectly capitalized forms appear as errors in proselint's output.

The actual result is that no errors are found and proselint returns nothing. 

I suspect this has to do with a bug in the tools.py function preferred_forms_check() but I'll have to investigate further. 

[pref_prose_test.txt](https://github.com/amperser/proselint/files/247111/pref_prose_test.txt)

KVGarg:
Hey @joshmgrant , if you don't mind, can I work upon this issue ?

PR #481, opened 2016-05-03
Unit tests based on possible bug with tools.py/preferred_forms_check.py

joshmgrant:
This is a unit test that may help confirm the issue raised in #480. Here, test_smoke() passes but test_capitalization() fails even when ignore_case is turned off.


suchow:
Thanks, I'm going to add a skip flag so that we can merge this in without breaking the build.

PR #482, opened 2016-05-09
Fixed spelling

TalkingAvocado:
Believe "femaile airman" to be a spelling mistake, though I haven't read Garner's Modern American Usage or encountered either version before so can't be sure.


suchow:
Good catch, thank you!

PR #484, opened 2016-05-16
Tests for garner.cliches

joshmgrant:
Here are some basic tests for the cliches module. Let me know what you think!


joshmgrant:
Based on the above failed checks, I'm not sure why the TravisCI checks failed. I don't understand the pep257 errors, but I'm able to reproduce them locally.

PR #486, opened 2016-06-09
Guidelines based on GLAAD Media Reference

joshmgrant:
June is Pride Month, so I thought I'd contribute in a fitting way by adding some checks based on [GLAAD's media reference](http://www.glaad.org/reference). These suggestions have been adopted by publications such as the AP and New York Times.


suchow:
Thanks, looking through this now, will make comments in a moment.

proselint/checks/glaad/terms.py
```diff
@@ -10,21 +10,31 @@
...
+
+    list = [
+        ["gay man",            ["homosexual man"]],
+        ["gay men",            ["homosexual men"]],
+        ["lesbian",            ["homosexual woman"]],
+        ["lesbians",            ["homosexual women"]],
+        ["gay people",            ["homosexual people"]],
+        ["gay couple",            ["homosexual couple"]],
+        ["sexual orientation",           ["sexual preference"]],
+        ["gay lives",            ["homosexual lifestyle", "gay lifestyle"]],
```
suchow: The suggested replacement probably won't work in may contexts:

"...the Party should regard participation in the self-destructive homosexual lifestyle as incompatible with..."

"Despite their explicit subject matter, the novels tended to include ham-fisted moral lessons about the dangers of the homosexual lifestyle;"

Is there another replacement that might work?

proselint/checks/glaad/terms.py
```diff
@@ -10,21 +10,31 @@
...
+        ["out",            ["admitted homosexual", "avowed homosexual"]],
```
suchow: This replacement has the same issue:

"The chairman of my Official Board was an admitted homosexual."

GLAAD recommends "openly gay", which I think would be better here. (This would also require changing the matched text from "admitted homosexual" to "an admitted homosexual".)

proselint/checks/glaad/offensive_terms.py
```diff
@@ -0,0 +1,38 @@
...
+    """Flag offensive words based on the GLAAD reference guide."""
+    err = "glaad.offensive_terms"
+    msg = "Offensive term. Remove it or consider the context."
+
+    list = [
+        "fag",
+        "faggot",
+        "dyke",
+        "homo",
+        "sodomite",
```
suchow: Can you change `ignore_case` below to false so that we don't flag `Sodomite`?

proselint/checks/glaad/offensive_terms.py
```diff
@@ -0,0 +1,38 @@
...
+def check(text):
+    """Flag offensive words based on the GLAAD reference guide."""
+    err = "glaad.offensive_terms"
+    msg = "Offensive term. Remove it or consider the context."
+
+    list = [
+        "fag",
+        "faggot",
+        "dyke",
+        "homo",
```
suchow: This is going to false alarm on "homo sapiens" and "homo erectus", so let's comment this particular word out (and open an issue?) until we have a topic-detector that can distinguish the two usages.

suchow:
This was a great idea, thank you for the contribution. Can you:

1. Rebase this onto master
2. Edit based on the comments.
3. Remove commits 1f2db7f and f8ec04e

Once that's done, it looks ready to merge!

proselint/checks/glaad/terms.py
joshmgrant: @suchow I've been thinking about this issue. It's tricky because the point is that there is no particular "homosexual lifestyle". I'm going to move these phrases to `offensive_terms.py' as terms to avoid in general as there's not really an obvious replacement.

suchow:
Hi Josh, two final requests:

- [ ] Can you rename this to `lgbt.terms`? We've moved from naming modules by their source to naming them by their contents. Ideally, we could find a word that meant "Prejudiced or insensitive language related to LGBT", like we did with `sexism.misc`, but that's not essential. 
- [ ] Can you tweak the spacing in `glaad.terms` so that the detected strings all line up like this?

```
list = [
    ["gay man",            ["homosexual man"]],
    ["gay men",            ["homosexual men"]],
    ["lesbian",            ["homosexual woman"]],
    ["lesbians",           ["homosexual women"]],
    ["gay people",         ["homosexual people"]],
    ["gay couple",         ["homosexual couple"]],
    ["sexual orientation", ["sexual preference"]],
    ["openly gay",         ["admitted homosexual", "avowed homosexual"]],
    ["equal rights",       ["special rights"]]
]
```

Thanks.

suchow:
This will appear in the next release, likely 0.6.2.

joshmgrant:
@suchow Great! I'm glad to see it go in. Thanks!

suchow:
I realize that we should have added this to the list in `README.md` and on the website. Eventually that will be automated, but for now it's done by hand. Maybe we need a checklist for new modules (or a linter? ha).

PR #488, opened 2016-06-11
Call load_options() once and save the results instead of calling it twice

kerkeslager:
Proselint starts up by reading and parsing the configuration files; due to the way this was structured it was reading them twice. I restructured slightly so that it would read them only once.


suchow:
Thank you! I'm going to rebase this onto the latest master and then merge.

PR #492, opened 2016-06-23
Fix typos

jwilk:
These misspellings were found using [mwic](http://jwilk.net/software/mwic).


suchow:
Cool project! Thanks for contributing.

PR #506, opened 2016-07-06
Change file opening to replace characters on UTF-8 decode errors

### LINKED ISSUE #504: proselint choked on bad character
my first attempt to use proselint choked on a bad character. (error message from macOS Terminal below). I have little experience with Python and Terminal, but it seems to be running. The plain text file was converted from Word by converting first to .odoc (because Word crashes when I ask for plain text), and then using macOS TextEdit to convert odoc to plain text. It seems unlikely that TextEdit would produce a bad text file. i attached the text file. 

Good news. I remembered that BBEdit has a zap gremlins command. After that, proselint worked. You might want to pass on that tip to others using proselint with BBEdit.

Thanks
Denis

p.s. The release notes for BBEdit mentions it's new command tool bbresults and gives a shout out to your proselint tool. This may encourage more users, like me, who are not familiar with Terminal. It would be nice if your READ ME offered a usage example, using bbresults, and explained that BBEdit has a menu command that results the path to the open file.

Denis Pelli
denis.pelli@nyu.edu

[Legibility87.txt](https://github.com/amperser/proselint/files/350720/Legibility87.txt)

Deniss-MBP-5K:~ denispelli$ proselint ~/Dropbox/legibility/Legibility87.txt | bbresults
Traceback (most recent call last):
  File "/Library/Python/2.7/site-packages/proselint/command_line.py", line 116, in proselint
    errors = lint(f, debug=debug)
  File "/Library/Python/2.7/site-packages/proselint/tools.py", line 216, in lint
    text = input_file.read()
  File "/System/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/codecs.py", line 302, in decode
    (result, consumed) = self._buffer_decode(data, self.errors, final)
UnicodeDecodeError: 'utf8' codec can't decode byte 0xd5 in position 1322: invalid continuation byte
Deniss-MBP-5K:~ denispelli$

CraigKelly:
Fixes #504 

I'm not sure what project owners would prefer to do when reading a file with invalid UTF-8 sequences, but here's a compromise (replace invalid characters on error).

I verified that the file attached in the issue failed on my box under both Python 2 and 3, and that the fix works under both Python 2 and 3 on the new file.

### DIFF
--- a/proselint/command_line.py
+++ b/proselint/command_line.py
@@ -112,7 +112,7 @@ def proselint(paths=None, version=None, clean=None, debug=None,
     num_errors = 0
     for fp in filepaths:
         try:
-            f = click.open_file(fp, 'r', encoding="utf-8")
+            f = click.open_file(fp, 'r', encoding="utf-8", errors="replace")
             errors = lint(f, debug=debug)
             num_errors += len(errors)
             print_errors(fp, errors, output_json, compact=compact)


suchow:
Thanks for your contribution, which has been rebased over master and merged via #514. I'm looking into adding invalid UTF-8 characters as a test case, but I'm having a bit of trouble, so I'll save that for later (unless you want to give it a go?).

CraigKelly:
I gave it a go - see PR #520 :)

PR #509, opened 2016-07-12
Update README.md with instructions on running tests

vikasgorur:
Is proselint in Python2 or 3? py.test seems to run fine for me with Python 3.5. Please let me know which is the correct version of Python developers are supposed to use and I'll update the README.

Thanks for this wonderful tool and I'm excited to contribute more!

### DIFF
--- a/README.md
+++ b/README.md
@@ -186,6 +186,44 @@ Interested in contributing to `proselint`? Great — there are plenty of ways yo
 - [Issue Tracker](http://github.com/amperser/proselint/issues)
 - [Source Code](http://github.com/amperser/proselint)
 
+### Development
+
+Proselint is written in Python3. Once you've cloned the repository and made local changes, you can run
+
+```bash
+$ python setup.py develop
+```
+
+to install your local copy.
+
+To run the tests, first install `py.test`:
+
+```bash
+$ pip install pytest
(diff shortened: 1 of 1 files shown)


michaelpacer:
Proselint should be both Python 2 and Python 3 compatible. 

My suggestion is aim at working in Python 3, and then work to bring back Python 2 compatibility. 

A good way to be able to try with both is to use a virtual environment manager toolset like `conda` or `virtualenv` + `virtualenvwrapper`.

vikasgorur:
Got it, thanks.

michaelpacer:
It's a good point though that explaining all of that just to run tests is a bit excessive for that part of the README. Can you think of any way to make sure tests are run for both without needing to first teach people how to use virtual environments?

vikasgorur:
I agree that it's too much to explain virutalenv. I just wrote your comment about doing development in Python 3 for now.

I think it's safe to assume that anyone who's willing to fix compatibility issues between Python 2 & 3 already knows how to use virtualenv.

vikasgorur:
Now that I've seen what's involved in getting tests to run with both Python 2 & 3, I think this PR is inadequate. I'll discard this and come up with something better.

Issue #345, opened 2016-03-08
Test system assumes proselint is installed instead of testing the code in the source tree


viccuad:
For example, https://github.com/amperser/proselint/blob/master/tests/_test_version.py#L17 

`out = subprocess.check_output("proselint --version", shell=True)`

suchow:
@viccuad thanks, good catch. would you be willing to take a stab at fixing this?

viccuad:
@suchow I'm afraid my Python-fu is not up to the task yet, so I would love if people chime in.

Nytelife26:
This was fixed in #1381 at the latest, if not much sooner when we were using poetry.

PR #512, opened 2016-07-12
Sort output by line and column number

### LINKED ISSUE #505: sort output by line number?
The output in my BBEdit results window is helpful and readable, but would be much easier to use if it were sorted by line number. here's an example:

~/Dropbox/legibility/Legibility87.txt:1339: after_the_deadline.redundancy Redundancy. Use 'bit' instead of 'tiny bit'.
~/Dropbox/legibility/Legibility87.txt:60: after_the_deadline.redundancy Redundancy. Use 'feelings' instead of 'inner feelings.'.
~/Dropbox/legibility/Legibility87.txt:1124: cliches.garner 'first and foremost,' is cliché.
~/Dropbox/legibility/Legibility87.txt:52: consistency.spacing Inconsistent spacing after period (1 vs. 2 spaces).
~/Dropbox/legibility/Legibility87.txt:57: consistency.spacing Inconsistent spacing after period (1 vs. 2 spaces).

vikasgorur:
Fixes #505

### DIFF
--- a/proselint/tools.py
+++ b/proselint/tools.py
@@ -200,7 +200,7 @@ def line_and_column(text, position):
     """Return the line number and column of a position in a string."""
     position_counter = 0
     for idx_line, line in enumerate(text.splitlines(True)):
-        if (position_counter + len(line.rstrip())) > position:
+        if (position_counter + len(line.rstrip())) >= position:
             return (idx_line, position - position_counter)
         else:
             position_counter += len(line)
@@ -235,7 +235,7 @@ def lint(input_file, debug=False):
             break
 
     # Sort the errors by line and column number.
-    errors = sorted(errors[:options["max_errors"]])
+    errors = sorted(errors[:options["max_errors"]], key=lambda e: (e[2], e[3]))
 
     return errors
 
--- a/tests/test_tools.py
+++ b/tests/test_tools.py
@@ -1,0 +1,35 @@
+"""Test the tools module."""
+
+from __future__ import absolute_import
+
+from .check import Check
+
+from proselint.tools import lint as lint
+
+
+class TestLint(Check):
+    """The test class for tools.lint."""
+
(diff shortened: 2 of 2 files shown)


tests/test_output.py
```diff
@@ -0,0 +1,15 @@
...
+from future.moves.subprocess import getoutput
+
+
+def test_output_sorted():
+    """Ensure that the output is sorted by line & column number."""
+    def extract_line_col(line):
+        pieces = line.split(":")
+        return (int(pieces[1]), int(pieces[2]))  # line number, column number
+
+    output = getoutput("proselint --demo")
```
vikasgorur: This test works fine locally but fails in CI. I suspect that in CI it's invoking an older installed version of `proselint` instead of the one defined by this revision.

suchow:
@vikasgorur I think you're right about the CI testing the installed version and not the one defined in the revision (see also #345 on this). I'm not sure how to fix it.

I'm wondering also if the right place to do this sorting might be in `proselint/tools.py`'s `lint` function, which the command line calls. That has the added benefit that the test can import `tools.lint` directly and bypass #345 until we figure it out.

vikasgorur:
@suchow yeah it looks like the `lint` function is already doing the sorting but doing it incorrectly. I'll just fix it there and add a test. Thanks!

suchow:
@vikasgorur Great, thanks for contributing!

suchow:
@vikasgorur This looks ready to merge, but first can you squash each stylistic fix into the preceding commit where that text or code was initially introduced?

vikasgorur:
@suchow I squashed it all into one commit since it's such a small change.

joshmgrant:
@vikasgorur Good work on getting that test coverage up. It's a huge leap!

vikasgorur:
Thanks!

PR #518, opened 2016-07-14
Add a helper script to run CI checks locally

vikasgorur:
I've added a script that runs the same commands that are run in CI. This should be useful to make sure a pull request will go through fine.

I also fixed a couple of bugs that were cluttering up the pep257 & test output.


suchow:
Thanks! This has been merged via #529.

PR #519, opened 2016-07-14
Add a check for misc. clichés.

vikasgorur:
I don't know what the policy on adding checks based on personal opinion is, but I hope we can agree that this is an atrocious cliché.

Please feel free to discard if the aim is to only add advice from published sources.

### DIFF
--- a/proselint/checks/cliches/misc.py
+++ b/proselint/checks/cliches/misc.py
@@ -853,3 +853,19 @@ def check_cliches_gnu_diction(text):
     ]
 
     return existence_check(text, list, err, msg, join=True, ignore_case=True)
+
+
+@memoize
+def check_cliches_misc(text):
+    """Check the text.
+
+    source:     various
+    """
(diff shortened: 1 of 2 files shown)


suchow:
Let's close this for now because, let's face it, inclusion requires an authority on English usage to have marked it as cliché. This could be a good addition to a `tics` module that I'm hoping to create; it looks for overuse of phrases like this one, "actually", etc.

vikasgorur:
Touché.

michaelpacer:
@vikasgorur If you wanted to, you could begin putting together such a list ("actually", "let's face it", "however") that would be a good start. 

@suchow do you have any sources on this list of words? I'm pretty sure there's a strunk and white rule that covers not using the same sentential construction again and again, but that would suggest a slightly different kind of check. Do you think that'd require a full phrase parsing? Could we use a prefix code as a way of making this kind of check efficient? Would we want it to go across documents?

PR #520, opened 2016-07-14
Testing for illegal utf-8 seq handling

CraigKelly:
Added a test asset (illegal-chars.txt) from the original issue (#504) and a test to insure the changes in the PR (#506) work correctly.


suchow:
Thanks! For reasons of copyright, I am concerned about including Denis Pelli's unpublished abstract as a test case within the Proselint source code. Can you reduce `illegal-chars.txt` to a minimal example that contains only the offending characters and perhaps, if you think it's useful, some text describing its contents (e.g., `This file contains an invalid UTF-8 character...`)? Then please rebase over master and I think we're ready to merge.

CraigKelly:
Sorry about that - I should have modified the file from the start. I just got a little excited. I believe the test is now working and correctly reading/interpreting output.

suchow:
Thanks!

PR #548, opened 2016-07-25
Add new rule: sentence spacing

Decagon:
Throw an error if two or more spaces are used `  ` to separate two words in a sentence from each other, or if there is extra space at the end of the sentence, provided that there is content after it, even if it's on the next line.


suchow:
Nice work, thanks for this. Can you also add a test?

Decagon:
I'll get some tests made up soon

michaelpacer:
This may need to wait until we have various code and other syntax specific lint-filters because I could see  this having huge issues with formatting (e.g., if something is a table in `.rst` all the columns need to be of the same width). It also might crop up in the case of `⇥` being converted to `␣␣␣␣` in python…

Also, I feel mixed on this one, but there is ambiguity in what counts as a false positive…and whether it applies in the plaintext or some formatted case. 
For example in LaTeX, if you were to put more than one space after a word in one case and not in another case, it actually would be interpreted correctly, because LaTeX just interprets it as a continuous string of whitepsace of arbitrary length (there might be exceptions to that…but it's generally true) and prints it as just a single space as defined by the overarching style of the document. 

So from the plaintext perspective, it's wrong. From the formatted document perspective, it's not wrong. So which of these is canonical? Would it count as a false positive? 

Pinging @suchow because this is something we already have run into with our single vs. double space after a period rule as regards LaTeX linting. 

So, @Decagon could you put together as part of the tests that you are generating a set of documents that would be likely to have lots of formatting specifics for different filetypes and formats? Because that would be what we'd need to be testing against (not just the New Yorker) to avoid false positives for rules like these.

Decagon:
@michaelpacer I was just thinking the same thing earlier today. If I restricted the rule to just too many spaces after a period, a comma, or a semi-colon (and don't worry about inter-sentence spacing) would that be a bit better? After investigating, this would cause a lot of the books that have centered titles to throw false alarms. Maybe we could disable it by default and only use it on formatted documents.

Decagon:
I'll close this for now because I have a bit of a mess on my master branch

suchow:
@Decagon To clean up a PR like this, you could rebase it over the master branch (`git rebase master`) and then force push (`git push --force`). Hm, maybe that's not what you meant by a mess though.

Decagon:
Ah, that seems to have worked. Unfortunately I can't reopen this because I forced push to the branch, maybe I did something wrong.

PR #556, opened 2016-07-26
Use the word 'very'

suchow:


suchow:
@danger really impressed by your tool, thanks!

PR #557, opened 2016-07-27
Fix #540

### LINKED ISSUE #540: Exclamation point warning on all caps text
I am using proselint 0.6.0 with python 3.5. I have the following sample text:

```
Hello, yes I AM A DOG.
proselint rules!!!!!
```

This produces the proselint warning:

```
$ proselint proselint_test.txt 
proselint_test.txt:1:11: leonard.exclamation.multiple Stop yelling. Keep your exclamation points under control.
proselint_test.txt:2:11: hyperbolic.misc 'rules!!!!!' is hyperbolic.
```
It seems like leonard.exclamation.multiple is catching multiple capitalized words rather than exclamation points, so it is confusing that it warns me about the latter.

Decagon:
This matches more than two exclamation marks, ignores capital letters, and some white space effectively fixing #540 .

### DIFF
--- a/proselint/checks/typography/exclamation.py
+++ b/proselint/checks/typography/exclamation.py
@@ -24,7 +24,7 @@ def check_repeated_exclamations(text):
     err = "leonard.exclamation.multiple"
     msg = u"Stop yelling. Keep your exclamation points under control."
 
-    regex = r"[^A-Z]\b((\s[A-Z]+){3,})"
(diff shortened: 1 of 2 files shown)


suchow:
Thanks. Can you add a test that includes the original bug or something similar?

Decagon:
Yeah I can add a test for that

suchow:
👍

PR #562, opened 2016-07-27
Add "kind of" and "mildly" to uncomparables

Decagon:
This commit is based from a user report:

> 'very unique' is caught, but not 'kind of unique' or 'mildly unique' or similar.


suchow:
Thanks! Looks mildly perfect to me.

PR #566, opened 2016-07-27
Replace "0" with "no" and "1" with "yes"

Decagon:
Zeros and ones can be ambiguous (well, not to us programmers) but to the general public.


suchow:
I'm going to close this because, though you make a good point about what would be readable to laypeople, this isn't meant for them. Instead, I've opened a new issue (#570) to create a CSV Schema, which is essentially a key for the formatting of CSV data.

Also, is decagongithub@gmail.com a good place to email you?

Decagon:
No problem. Yeah @suchow that's a good place to email me

PR #568, opened 2016-07-28
Add cats to list of venerys

### LINKED ISSUE #132: Add check for using the right name for groups of things
e.g., a group of cats is a "glaring".

http://www.oxforddictionaries.com/us/words/what-do-you-call-a-group-of

Decagon:


suchow:
🐈 

Good addition.

I don't think that merging of this PR should close #132, which is meant to have a much broader scope than the small list of terms currently included. The ultimate goal is for it to be complete. I'm not sure what the best way to handle this is — maybe you have thoughts?

Decagon:
I think it's a good idea to add some more animals to that list.

We can research some more venerys, and continue to add them until we have pretty much any animal that an average person would write about (since one can assume that people who are writing about very uncommon animals like the [The Blue Parrotfish](https://en.wikipedia.org/wiki/Blue_parrotfish) would be specialized in their area of research and so would use the correct venery.)

suchow:
@Decagon Sounds good. I'll merge this once the checks pass. The approach that @michaelpacer and I would like to take is to find some book or reference on venery terms, and then try to implement the entire book in code. I bet there's one out there.

michaelpacer:
This might be a good resource:
https://www.amazon.com/Exaltation-Larks-Ultimate-James-Lipton/dp/0140170960

PR #583, opened 2016-08-08
Test Class for `airlinese`

joshmgrant:
This is a basic test for the `airlinese.misc` check. It can be added to if desired.


joshmgrant:
No longer needed, can be removed.

PR #592, opened 2016-08-17
Ensure a clean commit history

suchow:


suchow:
@ProselintBot hurray, you're here.

PR #613, opened 2016-09-06
turn on the lgbtq checks

joshmgrant:
It appears the `lgbtq.offensive_terms` and `lgbtq.terms` modules are missing from the `.proselintrc` file. This adds them in and uses the checks from them by default.


suchow:
Can you also add this to the README and the website docs?

michaelpacer:
Note — links from danger do not resolve correctly when viewed in email.

PR #614, opened 2016-09-12
Make installation section more prominent

KrauseFx:
I scrolled through the documentation multiple times before seeing the `pip install` command, let's make it more prominent :+1:


suchow:
Thanks!

Issue #564, opened 2016-07-27
Detects gender bias despite gender neutral alternative term (from GetSiteControl)


alexyorke:
A user says:

> Detects gender bias on 'chairperson' despite the term being the gender neutral alternative to gendered nouns.

Text that will trigger false positive: `They thought the chairperson was interesting.`

suchow:
The fix here is to split up `sexism.misc.check` into two different check functions. The first flags gendered language; the second flags nonpreferred forms of gender-neutral language ("policeperson" vs. "police officer", "chairperson" vs. "chair").

michaelpacer:
Note, that in this case the "chairperson" vs. "chair" could be a false positive in the sense that it is introducing an error into the sentence, since now it is ambiguous as to whether it is referring to a chairperson or a physical chair.

suchow:
@michaelpacer ha, i think we can safely make the suggestion without needing to worry about that. "Bob Corker is the chair of the United States Senate Committee on Foreign Relations. He badly needs reupholstering."

Nytelife26:
This was resolved in #617.

PR #617, opened 2016-09-14
Split gender bias error messages to allow for preferred forms

kylesezhi:
As discussed in issue #564, split the gender bias misc into two parts: gendered language and preferred forms.


kylesezhi:
Code Climate thinks there is duplicate code in a list declaration - I don't think that's the case.

suchow:
Looks great, thanks!

PR #620, opened 2016-09-23
fix(checks): make am_pm check case sensitive

chreekat:
Upon using proselint on an email, I immediately noticed that "lowercase am/pm" was not checking that am/pm was actually lower case.

I attempted to fix that here (and throw in some tests). But, I'm not a python hacker, and I have no idea what the usual workflows are for python projects. I did run `nosetests` within the tests directory, but my new test still fails. Is there some build step I'm missing? Or is my test broken?

P.S. It would be great if you could throw a few words into the contributing guide about how to approach the project as an absolute Python beginner. E.g. is `pip install --user --upgrade .` the right way to install a modified version? Or, how do I run it without even installing it locally? :)

### DIFF
--- a/proselint/checks/dates_times/am_pm.py
+++ b/proselint/checks/dates_times/am_pm.py
@@ -19,7 +19,8 @@ def check_lowercase_periods(text):
     err = "dates_times.am_pm.lowercase_periods"
     msg = "With lowercase letters, the periods are standard."
 
-    return existence_check(text, [r"\d{1,2} ?[ap]m"], err, msg)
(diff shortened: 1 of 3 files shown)


chreekat:
Sorry for the close/reopen; I momentarily wanted nothing to do with this project after getting shit on in one of the issue threads.

Nytelife26 (approved):

Nytelife26:
Good fix, thanks for the PR - and I'm sorry for whatever happened in that issue all that time ago.

PR #625, opened 2016-10-14
Added weasel_words.misc check and test from stub

joshmgrant:
Looking at the test results the `weasel_words.misc` test was skipping because the check itself was still a stub. I decided to fill it in based on the source provided and enable it.


sunbubble:
Please merge this PR.

PR #633, opened 2016-11-15
executive secretary is not an oxymoron

j10sanders:
Executive secretaries work directly for and provide close administrative support to an executive. http://learn.org/articles/What_Does_an_Executive_Secretary_Do.html

I noticed the false positive while reviewing this article: https://www.firstthings.com/article/2016/10/the-genius-of-winding-paths

> Shortly after the Civil War broke out, Olmsted relinquished his position as park superintendent to become executive secretary to the Sanitary Commission.

I can imagine secretaries finding this offensive, and I see no reason why it should belong.

(Note, I am resubmitting this pull request because I removed the branch that originally had it.  sorry for duplicate!)

### DIFF
--- a/proselint/checks/oxymorons/misc.py
+++ b/proselint/checks/oxymorons/misc.py
@@ -36,7 +36,6 @@ def check(text):
         "build down",
         "conspicuous absence",
         "exact estimate",
-        "executive secretary",
         "found missing",
         "intense apathy",
         "mandatory choice",
@@ -44,6 +43,7 @@ def check(text):
         "organized mess",
         # "pretty ugly",
         # "sure bet",
+        # "executive secretary",
     ]
 
     return existence_check(text, oxymorons, err, msg, offset=1, join=True)


suchow:
"Executive secretary" is given as one of the examples of an oxymoron in Garner's Modern English Usage and so before writing it off as a false alarm, I'd like to figure out whether there are perhaps multiple uses of the term, some of which are fine, others of which are oxymorons.

Uses I can find:

1. The Sanitary Commission includes a job role called Executive Secretary, where "Secretary" means the same as it does in "Secretary of State". This role appears in many other commissions, too, though seems to be falling out of favor ([Google Ngram Viewer](https://books.google.com/ngrams/graph?content=executive+secretary&year_start=1800&year_end=2000&corpus=15&smoothing=3&share=&direct_url=t1%3B%2Cexecutive%20secretary%3B%2Cc0); see in particular the 1940s). Maybe Garner sees this job title as being an oxymoron.

2. When an executive at a company hires an administrative assistant, that assistant can be called an executive secretary. Perhaps "executive's secretary" would be better.

I'll merge this because the dominant usage appears to be #1 and it is no an error to call something by its name, even if that name were to contain a usage error.

I'll also send a tweet to Bryan Garner asking about this.

PR #634, opened 2016-11-21
Groups

j10sanders:
Fix issue 132


j10sanders:
The groups.py format is a bit unordinary because I made it possible to have more than one generic group term to look for (without duplicating code).

Currently, the generic terms are set to "group" and "bunch".  I imagine there are other generic grouping terms that people use too often, and they can easily be added to the `terms` variable.

joshmgrant:
@j10sanders Have you considered adding a test corresponding to `groups`? If it's a new module it would be good to have a small set of unit tests added with the new check module.

j10sanders:
@joshmgrant,  good call.  I've added tests to the pull request.

j10sanders:
@michaelpacer @suchow, any feedback on this pr?

j10sanders:
ping @michaelpacer @suchow

suchow:
Hi @j10sanders. Sorry for the delay in reviewing this PR. I think I may have misunderstood your comments on #132, which I had understood as a proposal to flesh out `terms.venery` [1], not create a new check. Is this meant as a replacement? Otherwise, it would be good to combine them.

[1] https://github.com/amperser/proselint/blob/master/proselint/checks/terms/venery.py

j10sanders:
Thanks, I hadn't realized that check existed.  I believe mine completely replaces it, but I'll check to be sure.

j10sanders:
@suchow, which format do you prefer?  Should I merge `groups` into `vennery`, or the other way around?

j10sanders:
@suchow, which format do you prefer? Should I merge `groups` into `vennery`, or the other way around?

suchow:
Hey, sorry for the slowness — let's bring everything into `terms.venery`.

Issue #624, opened 2016-10-13
-d option broken?


nbecker:
proselint -d outroute_timing_design.txt 
Deleting the cache...
rm: invalid option -- 'p'
Try 'rm --help' for more information.

 proselint -v
0.7.0

running on python-3.5.1

m-charlton:
It looks as if the problem is in the `clear_cache` function in 
`proselint/command_line.py`:

```python
45   def clear_cache():
46    """Delete the contents of the cache."""
47    click.echo("Deleting the cache...")
48    subprocess.call(["find", ".", "-name", "'*.pyc'", "-delete"])
49    subprocess.call([
50        "rm",
51        "-rfv",
52        "proselint/cache",
53        "&&",
54        "mkdir",
55        "-p",
56        os.path.join(os.path.expanduser("~"), ".proselint"),
57    ])
```

The `rm` command is executed in the subprocess call OK. For some reason the 
`-p` argument is being evaluated as part of the call to `rm` not `mkdir`.

One solution would be to divide the `rm` and `mkdir` operations into separate 
subprocess calls

```python
    result = subprocess.call(["rm", "-rfv", "proselint/cache"])

    if result == 0:
        subprocess.call(["mkdir", "-p", os.path.join(os.path.expanduser("~"), ".proselint")])
```

An alternative approach would be to use the existing `shutil.rmtree(...)` and
`os.mkdir(...)` calls:

```python

shutil.rmtree("proselint/cache", ignore_errors=False)

try:
    os.makedirs(os.path.join(os.path.expanduser("~"), ".proselint"))
except OSError as os_excp:
    # The exception can be thrown if `~/.proselint` already exists, in this 
    # case would we want to clean all caches bellow `~/.proselint`? Another
    # reason could be for access issues, though you'd hope the user would have 
    # access to their own home directory. The `pass` may not be a good idea here:
    pass
    
```

Questions:

- `shutil.rmtree` does not work with symbolic links to directories. Would this be a problem?
- If the `proselint/cache` is *not* removed then should the `~/.proselint` directory allowed to be created?
- See the comments in `except OSError` above. 

I'm quite prepared to fix this and add unit tests.

Nytelife26:
Fixed in #655.

PR #655, opened 2017-01-09
fix_clear_cache_#624

m-charlton:
Replaces call outs to OS commands using os agnostic equivalents in the std lib.
Previously the OS callouts were being incorrectly evaluated and causing the
removal of the old cache to fail.

### DIFF
--- a/proselint/command_line.py
+++ b/proselint/command_line.py
@@ -48,19 +48,32 @@ def clear_cache():
     click.echo("Deleting the cache...")
 
     # see issue #624
+    _delete_compiled_python_files()
+    _delete_cache()
+    _create_home_cache()
+
+
+def _delete_compiled_python_files():
+    """ Remove files with a 'pyc' extension """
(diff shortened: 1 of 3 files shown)


m-charlton:
Please find enclosed a fix for #624 '-d option broken?'. This replaces the call
out to the os commands `find`, `rm` and `mkdir` with os agnostic equivalents
in the standard library.

I'd like to add some unit tests for the above. This will involve the addition 
of the <https://pypi.python.org/pypi/mock> library. I can raise this in a 
separate issue.

suchow:
@m-charlton Hi Michael, thanks for the pull request, looks great. I'm happy to merge this now, but will wait for the addition of unit tests if you're still interested & willing to add those.

m-charlton:
I'll look at adding unit tests. As this area of code involves interacting with the file system, I'll 
investigate 'mocking' these tests.

Cheers

Michael.

m-charlton:
O.K see pep8 errors will fix

PR #671, opened 2017-02-20
Add verbose flag to broken links regexp

jwilk:
The regexp contains whitespace that was supposed to be ignored.


suchow:
Thanks!

PR #682, opened 2017-02-27
Add tests/illegal-chars.txt to MANIFEST.in

viccuad:
This will ship that file in the sdist and allow
test_illegal_chars.py to pass.


suchow:
Thanks

PR #683, opened 2017-02-27
Emacs Flycheck plugin packaging

### LINKED ISSUE #527: How to install proselint plugin on Emacs
I have installed `proselint` and it works on the terminal. But when I try to include `proselint` plugin code on my `init.el` Emacs does not initialise. Is it possible to give some guidance about the installation process?

mavit:
This scaffolding turns `flycheck-proselint.el` into an Emacs package, which will enable it to be included in package repositories such as [MELPA](https://melpa.org/) for easy installation.

This fixes #527.

Once this is done, I will submit the following recipe to MELPA to get the package included in their repository:
```elisp
(flycheck-proselint
 :fetcher github
 :repo "amperser/proselint"
 :files ("plugins/flycheck/flycheck-proselint.el"
         "README.md"
         "LICENSE.md"))
```

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,6 +6,10 @@ python:
 - '3.3'
 - '3.4'
 - '3.5'
+before_install:
+  - curl -fsSkL https://gist.github.com/rejeep/ebcd57c3af83b049833b/raw > emacs-version-manager-travis.sh && source ./emacs-version-manager-travis.sh
+  - evm install $EVM_EMACS --use --skip
+  - ( cd plugins/flycheck && cask )
 install:
 - pip install -r requirements.txt
 - gem install travis
@@ -17,6 +21,7 @@ install:
 - gem install kramdown
 - gem install danger
 - gem install danger-prose
+- ( cd plugins/flycheck && cask install )
 before_script:
 - travis lint -x --skip-completion-check
 script:
@@ -32,6 +37,8 @@ script:
     cd ..;
     wget --spider -e robots=off -w 1 -r -p http://www.proselint.com;
   fi
+- emacs --version
+- ( cd plugins/flycheck && cask exec buttercup -L . )
 after_success:
 - coveralls
 env:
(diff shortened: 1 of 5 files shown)


suchow:
@mavit Thanks. Do you know of any way to test or demonstrate that this works as intended? Maybe something that can be added to our test suite? I don't use Emacs or Flycheck and thus don't have any way of testing this out.

mavit:
I see that Flycheck itself runs tests in Travis using a testing framework called buttercup, so it's certainly possible.  We can't copy and paste Flycheck's harness, however, since it's GNU GPL licensed.

I'll have a look into it when I find time.  Would you be satisfied with testing against a single version of Emacs, rather than the Cartesian product of Emacs and Python versions?

m-cat:
Any update on this? Would be very convenient if this were available as a package through melpa.

mavit:
Is there reluctance to merge this, or have you simply not had time to look at it?  Would it be better for me to split the Flycheck plugin out into it's own repository, I wonder?

suchow:
Done, thanks for the contribution! And quite sorry for the delay.

robstewart57:
Is there a reason why flycheck-proselint isn't listed as a melpa package?

https://melpa.org/#/?q=proselint

mavit:
Support for `proselint` is now built in to Flycheck, so no plugin is required.  https://github.com/flycheck/flycheck/pull/1304

robstewart57:
Excellent! So when I enable flycheck-mode, proselint is enabled by default if my computer has a proselint executable in `$PATH`? Or do I manually have to enable proselint checks when running flycheck against a particular buffer?

mavit:
You need to tell Flycheck to which modes is should apply `proselint`, if that mode is not [included by default](https://github.com/flycheck/flycheck/blob/master/flycheck.el#L8674).  For example:
```lisp
(flycheck-add-mode 'proselint 'org-mode)
```

But this is not a support forum, so such questions are better directed to http://www.flycheck.org/en/latest/community/get-help.html

PR #684, opened 2017-02-28
readme: Add ALE Vim plugin

polyzen:


suchow:
Thanks! Just to make sure I understand: if someone installs both ALE and has Proselint, there's no extra setup needed to get Proselint running in Vim? Or is there a Proselint plugin for ALE that they need to install?

polyzen:
No extra setup needed in general. The text filetype is disabled by default as that's also used for files with unknown filetypes. The author is mulling over whether it should be disabled by default for the Vim help filetype(/:help windows) as well.

polyzen:
I might be off by saying "text" is used for unknown filetypes, but here are the relevant threads:
w0rp/ale/issues/229
w0rp/ale/pull/367

PR #685, opened 2017-03-02
Fix broken links regexp for Python 2

### LINKED ISSUE #672: Bad escapes in broken links regexp
`proselint/checks/links/broken.py` contains the following code (with boring parts of the regexp omitted):
```python
    regex = re.compile(
        r"""...boring...
        |[^\s`!()\[\]{};:\'".,<>?\xab\xbb\u201c\u201d\u2018\u2019\u21a9]))""",
        re.U)
```
But the `\uXXXX` escape sequences for regexps were added only in Python 3.3.
In earlier versions, `\u` stands for literal `u`.
In other words, in Python 2.7, this code is equivalent to:
```python
    regex = re.compile(
        r"""...boring...
        |[^\s`!()\[\]{};:\'".,<>?\xab\xbb01289acdu]))""",
        re.U)
```
...which is certainly not what you wanted.

The dubious regexp was found using [pydiatra](https://github.com/jwilk/pydiatra).

jwilk:
* Make the regexp pattern a Unicode string.

* Don't use `\uXXXX` escape sequences; they are supported in regexps only since Python 3.3.

Fixes #672

(This is pretty ugly, but it's the best I could come up with.)

### DIFF
--- a/proselint/checks/links/broken.py
+++ b/proselint/checks/links/broken.py
@@ -23,6 +23,15 @@ except ImportError:
 from socket import error as SocketError
 standard_library.install_aliases()
 
+if str is bytes:
+    # Python 3 doesn't have ur'...' strings.
+    # With the help of this function, you can write u(r'...') instead.
+    def u(s):
+        return s.decode('utf-8')
+else:
+    def u(s):
+        return s
+
 
 @memoize
 def check(text):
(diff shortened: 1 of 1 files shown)


jwilk:
Closing in favor of #731.

PR #690, opened 2017-03-13
Fix one-off column offset with existence_check

blueyed:
When require_padding was not used (or it is at the beginning of a line),
columns were off by 1.

This patch adds/substracts the matched subgroup conditionally.

This could use more tests in general - none was failing with initial (wrong) changes in this area.

I've tried to use `\b` in the pattern instead, but that failed for/with `proselint/checks/misc/currency.py`.

### DIFF
--- a/proselint/tools.py
+++ b/proselint/tools.py
@@ -320,7 +320,7 @@ def existence_check(text, list, err, msg, ignore_case=True,
         flags = flags | re.DOTALL
 
     if require_padding:
-        regex = u"(?:^|\W){}[\W$]"
+        regex = u"(^|\W){}([\W$])"
     else:
         regex = u"{}"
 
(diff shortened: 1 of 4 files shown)


blueyed:
Running into `AssertionError: sorry, but this version only supports 100 named groups` with this approach (except on py35).

joshmgrant:
@blueyed looks like the issue is coming from `tools.py` in the line compiling the regex causing your failed tests The ()'s here are creating too many groups for compiling regexes. Maybe use []'s here instead?

blueyed:
@joshmgrant 
Thanks.
I assume it's from https://github.com/amperser/proselint/pull/690/files#diff-a2bedf23c7c2dfe9ca9c71e77c10500cR323, and then they are needed, since they are used to adjust the offset - based on if `\W` or `^` was matched.
I've not put much thought into it, I was just happy to see that it's easily hackable when finding the issue itself.. ;)

Nytelife26:
Can anyone verify if this problem still occurs? I am currently not able to, so any help would be appreciated.

Nytelife26:
This was resolved in #1399 at the latest.

Issue #605, opened 2016-08-28
Dubious: Starting paragraphs with 'But'


cpitclaudel:
Is there a source for the following warning?

```
misc.but No paragraph should start with a 'But'.
```

Here's what Garner says on the subject (but, 3rd edition, p 121):

> It is a gross canard that beginning a sentence with _but_ is stylistically slipshod. In fact, doing so is desirable in any number of contexts, as countless stylebooks have said.

mcepl:
Moreover, there seems to be some problems with parsing (at least in
``rST``):

```rst
Ideological struggle
    In the conditions of freedom it is just a contest of ideas,
    but in the militant ideological state it means pushing of one
    ideology and suppressing all others. Weakness: what is
    imposed generates resistance.
```

Whatever one thinks about this particular problem, there is no start of the paragraph here.

Also, this is effectively a duplicate of the discussion on #98.

sobjornstad:
I don't think I've ever tried to start a paragraph with a “but” (though I have no opinion on whether it should be considered wrong: sentences are fine, why not paragraphs?). But (haha) when using [semantic linebreaks](https://sembr.org) or indeed any kind of hard line breaks in your text, this rule yields false positives all over the place...it seems to be checking the beginning of *lines* instead of the beginning of paragraphs.

Nytelife26:
> Here's what Garner says on the subject (but, 3rd edition, p 121):
> 
> > It is a gross canard that beginning a sentence with _but_ is stylistically slipshod. In fact, doing so is desirable in any number of contexts, as countless stylebooks have said.

That is referring to sentences, not paragraphs. Do note that in the [original issue] that lead to the creation of this rule, the subject of paragraphs and self-containment is addressed.

[original issue]: https://github.com/amperser/proselint/issues/98

sanjayankur31:
> Moreover, there seems to be some problems with parsing (at least in `rST`):
> 
> ```
> Ideological struggle
>     In the conditions of freedom it is just a contest of ideas,
>     but in the militant ideological state it means pushing of one
>     ideology and suppressing all others. Weakness: what is
>     imposed generates resistance.
> ```
> 
> Whatever one thinks about this particular problem, there is no start of the paragraph here.


This also affects TeX/LaTeX. I filed a new issue for this: https://github.com/amperser/proselint/issues/1376

Nytelife26:
[...]

PR #691, opened 2017-03-14
fix(checks): ignore case in misc.but for lowercase "but"

### LINKED ISSUE #98: Add rule about starting paragraph with "But"

### LINKED ISSUE #745: misc.but reports false positive
Run proselint on the following file:

```
This line is OK,
but this is not.
```

The following error is reported:

```
misc.but No paragraph should start with a 'But'.
```

That "but" is not only not at the beginning of the paragraph, it's not even at the beginning of a sentence.

blueyed:
Resolves #98. Resolves #745.

### DIFF
--- a/proselint/checks/misc/but.py
+++ b/proselint/checks/misc/but.py
@@ -22,4 +22,5 @@ def check(text):
     err = "misc.but"
     msg = u"No paragraph should start with a 'But'."
     regex = r"(^|([\n\r]+))(\s*)But"
-    return existence_check(text, [regex], err, msg)
+
+    return existence_check(text, [regex], err, msg, ignore_case=False)
--- a/tests/test_misc_but.py
+++ b/tests/test_misc_but.py
@@ -20,3 +20,5 @@ class TestCheck(Check):
         """Basic smoke test for misc.but."""
         assert self.passes("""Smoke phrase with nothing flagged.""")
         assert not self.passes("""But I never start with the word "but".""")
+        assert self.passes("""I never start with the word "but",
+but might use it after a linebreak.""")


KVGarg:
Fixes Issue #98 , #605

KVGarg:
Hey @blueyed , i was going through your changes and found an issue,
```
Hey i am testing

on but keyword,
but, wait, line four is wrong.
but, wait, why this line is not giving any warning?
```
last line should give a warning/error that the paragraph should not start with `but`.  I know that `but` should be in Caps as it's a new line but that is not a problem as i have added a check for that to raise warning whether sentence begins with Caps letter or not. Can you just modify the `regex` to solve this issue?

KVGarg:
Fixes Issue #98 , #605, #745

blueyed:
@KVGarg 
Thanks for triaging.
I assume the previous comment (I've received via email) is not valid anymore?

KVGarg:
Which comment ? because when I referenced other issues there were no other comments.

blueyed:
Deleted it already, but it was about multiple "but"s and the last one not being caught, asking to change the regexp.  Might have been lost during the GitHub outage yesterday then maybe (i.e. emailed, but not posted here)?!

KVGarg:
Ok @blueyed

Nytelife26 (approved):

suchow (approved):

PR #694, opened 2017-03-18
Remove unnecessary #! lines

mavit:
`rpmlint` is unhappy because these files are not executable but contain `#!` lines suggesting that they should be.


suchow:
Thanks!

Issue #704, opened 2017-04-27
Do not install tests in generic "tests" packages


languitar:
Please avoid installing the unit tests of this project into the very generic python package "tests". This is very likely going to result in name clashes.

joshmgrant:
@languitar Can you elaborate on this issue? I'd like to learn how I can help avoid any issues running the included tests in this project.

languitar:
In the current form, your `setup.py` includes the `tests` folder in the packages. As a result you might end up with something like this in the filesystem after doing `python setup.py install`:

```
(env) languitar@jaco ~/src/proselint (master)> 
ll env/lib/python3.6/site-packages/proselint-0.8.0-py3.6.egg/
total 12K
drwxr-xr-x 2 languitar languitar 4.0K Apr 27 14:56 EGG-INFO/
drwxr-xr-x 4 languitar languitar 4.0K Apr 27 14:56 proselint/
drwxr-xr-x 3 languitar languitar 4.0K Apr 27 14:56 tests/
```

Since `tests` contains an `__init__.py`, you can afterwards do:

```
>>> import tests
>>> tests
<module 'tests' from '/home/languitar/src/proselint/tests/__init__.py'>
>>> import tests.test_tools
>>> tests.test_tools
<module 'tests.test_tools' from '/home/languitar/src/proselint/tests/test_tools.py'>
```

That means you have made available your unit tests in the generic python namespace und a name that is very likely to collide.

The origin of this problem is very likely your plain call to `find_packages` in `setup.py`, which picks up every package - so the `tests` package as well`. There are [solutions on stackoverflow](https://stackoverflow.com/questions/26545668/setup-py-sdist-exclude-packages-in-subdirectory) on how to solve this issue.

joshmgrant:
Wow, great catch! This is likely a fairly easy fix but for now, I'd suggest folks _not_ import any form of tests from the project, so don't do any sort of `from proselint import tests` or similar if it can be avoided.

languitar:
Btw, I noticed this because of an explicit `rm` in the [archlinux user package for proselint](https://aur.archlinux.org/cgit/aur.git/tree/PKGBUILD?h=python-proselint).

suchow:
Closed by #706.

PR #706, opened 2017-04-28
small update find_packages() to exclude tests directory

joshmgrant:
Based on #704 I've made an update to `find_packages()` in setup.py to exclude all mention of tests.


suchow:
Thanks, Josh!

Closes #704.

PR #730, opened 2017-06-01
Fix indentation in proselintrc

jwilk:


Nytelife26:
Could not push to fork, so will have to reject.

PR #731, opened 2017-06-01
Enable unicode_literals for checks.links.broken

### LINKED ISSUE #672: Bad escapes in broken links regexp

jwilk:
This makes `\uXXXX` escape sequences in the regular expression work in Python 2.

Fixes #672

### DIFF
--- a/proselint/checks/links/broken.py
+++ b/proselint/checks/links/broken.py
@@ -13,6 +13,7 @@ categories: writing
 Check that links are not broken.
 
 """
+from __future__ import unicode_literals
 from proselint.tools import memoize
 from future import standard_library
 import re


Nytelife26:
As of [a study conducted by JetBrains in 2021](https://lp.jetbrains.com/python-developers-survey-2021/#PythonVersions), Python 2 now makes up for less than 5% of all usage. Since we are considering dropping `future` altogether, I feel confident in closing this.

PR #746, opened 2017-09-02
Proselint support will be built-in to Flycheck 31.

mavit:
Support for Proselint is now present in the upstream Flycheck (https://github.com/flycheck/flycheck/pull/1304).  Therefore, Proselint's own Flycheck plugin is now redundant.


suchow:
Thanks!

PR #750, opened 2017-10-03
Add Python 3.6, drop unsupported 3.3

hugovk:
## Reasons for dropping 3.3

* https://en.wikipedia.org/wiki/CPython#Version_history
* pip 10 will deprecate Python 3.3 support, pip 11 won't support it (https://github.com/pypa/pip/issues/3796)
* Very little PyPI traffic (June 2016) https://github.com/pypa/pip/issues/3796

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,6 +5,7 @@ python:
 - '2.7'
 - '3.4'
 - '3.5'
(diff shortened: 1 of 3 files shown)


suchow:
Thanks!

Issue #247, opened 2015-10-20
Extract rules on ableism


suchow:
http://www.autistichoya.com/p/ableist-words-and-terms-to-avoid.html

Andrew-Suber:
Reading this article, it is clear that there can be specific individual slurs, but many of the offenses are subtle and not considered universally offensive. For instance, 'burn victim' is not considered offensive by all.

Can we agree to perhaps flag the most offensive, recognizable slurs and shelve the rest of the issues related to Ableism?

PR #753, opened 2017-10-15
Ableism

ericwbailey:
Hi there,

This PR addresses https://github.com/amperser/proselint/issues/247, adding ableist terms from the link in the issue, supplemented with terminology from [this Wikipedia article](https://en.wikipedia.org/wiki/List_of_disability-related_terms_with_negative_connotations). 

I modeled the rules after the [LGBTQ checks](https://github.com/amperser/proselint/tree/master/proselint/checks/lgbtq), as it seemed to be the closest match to what I was going for: errors for outright slurs and warnings for potentially problematic language.

A few notes:

- I left terms repeated in other checks in, I figured it would be good to preserve context and intent for that specific check
- I had my friend @danielsmc take a final look before submitting, as I'm not too familiar with Python
- There's some initial tests included as well, I tried to model them after the other tests I saw in the repo

As an aside, I just wanted to say thanks for this project. It's really great to have as a resource, and I try to recommend it to everyone I know who is into linting!


joshmgrant:
@ericwbailey Great to see someone looking at my work 👍 

Your PR is currently not meragable due to some [pep8](https://www.python.org/dev/peps/pep-0008/) failures. You can see the specific ones in Travis-ci failed build above. Let me know if you need any help!

ericwbailey:
@joshmgrant Just pushed up some fixes, looks like the failures are all set now 🎉

suchow:
Thanks Eric, this is really excellent.

Before merging this, we should think about which of these will produce too many false alarms, and perhaps add some thresholding mechanism so that we don't flag perfectly fine language. (For example, a lone instance of `I like to eat macadamia nuts` will get flagged by the current implementation because it has the word `nuts`). I'll make specific line comments about particular words soon.

ericwbailey:
Yeah, definitely. Saw some code in the source about programmatically sussing out context, but wasn't sure if it was applicable here? I figured having a comprehensive list to whittle down would be preferable than something half-baked.

Anyways, looking forward to your comments!

PR #782, opened 2018-03-12
Make proselint compliant XDG spec

### LINKED ISSUE #747: Please use XDG paths for config and cache
The [XDG Base Dir Spec](https://standards.freedesktop.org/basedir-spec/basedir-spec-latest.html) is more than a decade old. It's discouraged to put app's configuration and cache in user's home. Please use shell envs `$XDG_CACHE_HOME` (fallback to `$HOME/.cache` if not set)  and `$XDG_CONFIG_HOME` (fallback to `$HOME/.config` if not set).

It's worth mentioning that this spec applies to cli tools too. For example git, systemd, htop, mc and many others honor it.

Please keep our homes clean, thanks!

xu-cheng:
The configuration file is now set as `$XDG_CONFIG_HOME/proselint/config`.
If `$XDG_CONFIG_HOME` is not set or empty, `~/.config/proselint/config`
will be used. Additionally for compatible reason, the legacy configuration
`~/.proselintrc` will be used if `$XDG_CONFIG_HOME/proselint/config`
does not exist.

Similarly, the cache location is set as `$XDG_CACHE_HOME/proselint`.
If `$XDG_CACHE_HOME` is not set or empty, `~/.cache/proselint` will
be used. It will also automatically migrate the legacy cache
`~/.proselint` to the new location.

Finally, remove the function `_create_home_cache()`, which is used in
`proselint --clean`, since the cache directory will be automatically
created and we have changed the cache location.

Fixes #747.

### DIFF
--- a/README.md
+++ b/README.md
@@ -123,7 +123,8 @@ This will return a list of suggestions:
 
 ### Checks
 
-You can disable any of the checks by modifying `~/.proselintrc`:
+You can disable any of the checks by modifying `$XDG_CONFIG_HOME/proselint/config`. If `$XDG_CONFIG_HOME` is not set or empty, `~/.config/proselint/config` will be used.
+Additionally for compatible reason, the legacy configuration `~/.proselintrc` will be used if `$XDG_CONFIG_HOME/proselint/config` does not exist.
 
 ```json
 {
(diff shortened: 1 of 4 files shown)


xu-cheng:
Any comment? Any chance to merge this?

suchow:
Thank you!

Issue #277, opened 2016-02-12
interaction with LaTeX?


nbecker:
Does proselint grok LaTeX?  I'm guessing for now converting LaTeX -> plain txt is best?

suchow:
It doesn't yet handle LaTeX source code in any special way, but that won't stop it from running over the document. You may find that it falsely flags some of the markup (e.g., straight quotes in image tags), but otherwise it should be fine.

j6k4m8:
I'm interested in extending this to run in LaTeX .tex documents. Do you have a preference for how/where I implement? Is this a `plugin`?

nbecker:
I have no preference

suchow:
@j6k4m8 The narrowest way to handle this (recommended) is to create a new topic detector and then exclude that new topic for all the checks that false alarm on LaTeX syntax. The ultimate solution is to parse the text from the LaTeX and apply Proselint to that parsed representation.

https://github.com/amperser/proselint/blob/master/proselint/tools.py#L433

j6k4m8:
@suchow Got it. My plan was to parse the LaTeX, keeping track of line-number so that the output could still reflect the original line-numbers and character positions. I imagine this would require modification to:

| Code | Description |
|-------|------------|
| [`main.`CLI args](https://github.com/amperser/proselint/blob/master/proselint/command_line.py#L113) | Adding a `--latex` arg (unless you want to infer from the text or file ext?)
| [`tools.line_and_column`](https://github.com/amperser/proselint/blob/master/proselint/tools.py#L198) | Modification to enable line number "hints" instead of using the enumerated line number from iterating over the file |

With a super basic python script to strip out LaTeX `\keywords{}`, such as this...

```python
import sys
import re


for i, line in enumerate(sys.stdin):
    i += 1
    line = re.sub(re.compile(r"\\\w+\{(.*?)\}"), r"\1\n", line)

    line = [n.strip() for n in line.split("\n") if n.strip()]
    for l in line:
        print("{}\t{}".format(i, l))
```

...I think we can have a working model pretty quickly. I appreciate there are [already tools that do this](https://github.com/pkubowicz/opendetex) but I'm not sure to what extent we can rely on them, and I'm not sure how they perform with the issue I mention below.

I think it's important to strip these tags out and parse the split text separately because of how commonly folks include bracketed section headers in the text content:

```latex
\section{Section Title} Here is the content of my
section. As you can see, lorem ipsum dolor sit amet...
```

[...]

PR #795, opened 2018-05-13
Don't match multiple dashes in check_ly

ekmartin:
In LaTeX, multiple dashes is used to represent an en (--) or an em (---) dash. Currently `check_ly` matches sentences on the form: `This is a family---albeit a small one (...)` which it probably shouldn't.


Nytelife26:
This would be better fixed with LaTeX and Markdown awareness. Relevant issue: #277 

I'll close this for further discussion there.

PR #797, opened 2018-05-29
added conditional requirement

joshmgrant:
It looks like openSUSE sometimes does not include `python-dbm` with standard Python 3 modules. This includes it if needed on all `linux` OSs.


suchow:
Thank you. Merging via #856 in a moment.

PR #802, opened 2018-06-13
Simple regex checker for passive voice constructions

RyanMcCarl:
This is my first attempt to contribute, so please let me know if I need to do anything differently -- e.g. register the check with the application.

I haven't tested the regex extensively, but it performed well on [this list](https://github.com/amperser/proselint/files/2096770/passive_voice_examples.txt).

Here are the results on [Pythex](https://pythex.org/):

[Passive voice regex test 2018-06-12.pdf](https://github.com/amperser/proselint/files/2096776/Passive.voice.regex.test.2018-06-12.pdf)

Here is the regex itself -- I'm sure it can be improved, modularized into several regexes, etc. Happy to continue working on it when I can if you like the idea.

```
(\b(?:be|am|is|are|was|were|have|has|had)\b[\w\s]{,15}?(?:d|(?<!whe)n|ne|left|being)\b(?: by\b)?)
```


RyanMcCarl:
I'm having trouble getting this through the checks even after rebasing the repository. Can anyone tell me what I'm doing wrong? @mpacer @suchow

Nytelife26:
It seems that for the most part this PR is, well, broken. Its contents appear to be lost in over 372 changed files. Using regexes to hack together a passive voice check also doesn't seem ideal.

With that being said, I think we'll have to close this, unless you're available to work on it @RyanMcCarl.

suchow:
@Nytelife26 Even if Ryan can’t work on this, it’s worth salvaging what we can from the PR. There are new modules buried in here based on Ryan’s professional expertise (he’s a lawyer and scholar who has taught advanced legal writing courses and does research on use of A.I. in law).

suchow:
In particular, see proselint/checks/mccarl/rm_style_pref_forms.py

Nytelife26:
Ah, I see. I may have to root through the diffs and cherry-pick useful files, then - I didn't have time to review them all due to the mostly broken structure and the number of file changes.

I'll take a look now, though.

Nytelife26:
I will review the Garner checks now as it is likely they are already present in another check since we have most of GMEU implemented. The McCarl checks and passive voice check will additionally need to be split up into two separate PRs before this can continue.

Nytelife26:
The regex needs to be adjusted to pass the special cases checks.

PR #832, opened 2018-07-08
update README and requirements

joshmgrant:
Nose is no longer really being used as a test runner, so this updates the docs and automated tests to use py.test instead.


suchow:
Thanks! Merging this commit in a moment via #840.

PR #835, opened 2018-07-18
Move semicolon out of quotes in weasel_words/very

cpitclaudel:
See e.g. https://owl.english.purdue.edu/owl/resource/577/03/


suchow:
Thanks! 

Closed in favor of PR #837, which has cpitclaudel's commit and will be merged in a moment.

PR #856, opened 2018-07-22
Add conditional requirement

suchow:
Merging #797.


suchow:
@joshmgrant Could you take another look at this? Getting build failures, something about the `sys_platform` not being recognized.

jayvdb:
This wont work consistently, as https://pypi.org/project/python-dbm/ is 404 , so lots of tools will not be able to resolve this requirement.

And also lots of in-use versions of setuptools/pip/packaging/etc which had problems with `sys_platform`.  Search online for setuptools+markers+sys_platform to get a feel for that.

It is now at https://build.opensuse.org/package/show/devel:languages:python/python-proselint and has the necessary conditionals there.

IMO this should be closed.

PR #869, opened 2018-08-02
Use ignore_case=False in dates_times.am_pm.lowercase_periods

io12:


suchow:
@io12 Thanks. I'm not sure that I understand this one — the advice is that, when you write a.m. or p.m. using lowercase letters, you use periods. Wouldn't ignoring case here give that advice when the writer uses capital letters, too?

Issue #800, opened 2018-06-03
Add checks for capitalizing the beginning of sentences.


Gigahawk:
Will require some kind of function that splits the given text into individual sentences, NLTK can do this but that's probably overkill and I was thinking of just taking [this](https://stackoverflow.com/a/31505798) and putting it in `tools.py`.

The check would just take the output of that function and return a warning for every line where the first alphabetic character isn't capitalized.

KVGarg:
@Gigahawk I would like to work upon this issue, can i ? and i think it should be added in `checks` folder by creating a new folder named `capitalize` instead of adding directly in `tools.py` because we are checking whether there are any issues in text or not.

Gigahawk:
Yea go ahead, I never really got anywhere with this.
I don't think I can assign issues but you can probably just reference this issue in your pull request.

KVGarg:
Okay @Gigahawk , I will reference this issue in PR

Nytelife26:
It is worth noting that this is solely a grammatical issue, entirely unrelated to prose in my mind. I am open to further discussion, but this seems to be outside of the scope of the project.

Gigahawk:
It's pretty typical for linters to catch syntax as well as style issues.

Nytelife26:
> It's pretty typical for linters to catch syntax as well as style issues.

While this is true, `proselint`'s sole focus is style. Our very explicit mission is to avoid becoming yet another grammar checker. Unless there is stylistic reasoning for this rule, I would be hesitant to add it.

Nytelife26:
Without further correspondence or justification, I will close this for now.

PR #893, opened 2018-10-17
Added checks for capitalizing the beginning of sentences

KVGarg:
1. Fixes Issue #800
2. Added a rule( named `split_into_sentences`) in `tools.py` that splits the text into sentences and return all the sentences.
3. Added check in `misc.capitalization` for checking the beginning of sentences.


KVGarg:
Will open PR with some modification in function `split_into_sentences` in `tools.py` that I have inserted.
As Websites were also getting split into new sentences so to avoid them adding some changes in function

PR #951, opened 2019-04-28
Allow including / excluding individual checks by key

### LINKED ISSUE #491: .proselintrc usage
Are there any guidelines on how to use `.proselintrc`?

I've found a `.proselintrc` in  `/usr/lib/python2.7/site-packages/proselint` and can modify the checks there, but it doesn't seem like I can override it with something like `~/.proselintrc`.

Also, is there any way to disable a particular rule? I want to stop curly quote checking, but the following doesn't work in `.proselintrc`.

```
...
"typography.symbols"            : true,
"typography.symbols.curly_quotes" : false,
...
```

Is there any way to keep the rest of the checks in `typography.symbols`, but disable one rule?

### LINKED ISSUE #729: Support more than two keys in .proselintrc
In #491, @netflash pointed out that:

> Looks like the filter in the tools.py only works with first 2 keys,
`"typography.symbols.curly_quotes": false,` won't work, and this
`"typography.symbols": false,` should work

It would be great if we could have more control over what rules to ignore.

### LINKED ISSUE #1091: Allow disabling individual checks

AlexRiina:
This pull requests allows the end user to include or exclude individual tests as part of the config file. I was using this to lint a file where my markdown pipeline automatically converted ellipsis and quotes to their nicer html entity equivalents. Prior to this pull request, I could only disable the entire typography.symbols suite but my pipeline would not automatically convert a `(c)` so I didn't want to disable the whole suite.

Resolves #729. Resolves #491. Resolves #1091.

### DIFF
--- a/README.md
+++ b/README.md
@@ -147,10 +147,13 @@ This will return a list of suggestions:
 You can disable any of the checks by modifying `$XDG_CONFIG_HOME/proselint/config`. If `$XDG_CONFIG_HOME` is not set or empty, `~/.config/proselint/config` will be used.
 Additionally for compatible reason, the legacy configuration `~/.proselintrc` will be used if `$XDG_CONFIG_HOME/proselint/config` does not exist.
 
+You can disable either entire sets of tests or individual tests from within a set through the config file.
+
 ```json
 {
   "checks": {
-    "typography.diacritical_marks": false
+    "typography.diacritical_marks": false,
+    "typography.symbol.ellipsis": false
   }
 }
 ```
--- a/proselint/tools.py
+++ b/proselint/tools.py
@@ -151,14 +151,31 @@ def get_checks(options):
     """Extract the checks."""
     sys.path.append(proselint_path)
     checks = []
-    check_names = [key for (key, val)
-                   in list(options["checks"].items()) if val]
 
-    for check_name in check_names:
-        module = importlib.import_module("checks." + check_name)
-        for d in dir(module):
-            if re.match("check", d):
-                checks.append(getattr(module, d))
+    # includes and excludes should be either modules like 'typography.symbol'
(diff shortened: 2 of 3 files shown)


Nytelife26:
This is a very interesting proposition. At present the config does only allow enabling or disabling modules, rather than specific sub-checks. How do we feel about this, @suchow?

suchow:
First, there are three interrelated issues here:

1. Proselint isn't Markdown-aware and so complains about syntax that's acceptable in Markdown but not in prose (e.g., `...` vs. `…`). To sidestep this issue, @AlexRiina wanted to disable only some of the typographic checks. But we should also fix the underlying problem. I think there's an issue open for this already, somewhere; people run into it all the time.

2a. There is no mechanism to disable only some of a module's checks.

2b. If we create a mechanism to disable only some of a module's checks using the proposed approach, it's quality will depend on how well we are using the check namespace, breaking modules down into logical submodules that people would want to independently turn on and off.

I'm 👍 on allowing enabling/disabling submodules via the config and think it'll be useful to many people. I'm agnostic as to whether the current PR is the "right" way to do that; I defer to you. I think we should, perhaps as part of this PR or as a follow-on PR, review the whole namespace and see if there's anything obvious that should be renamed, split, or combined.

Nytelife26:
It seems like that largely depends on namespacing, then. I believe this
should be resolved *after* our categorization evaluation, rather than
before. Therefore, we should postpone this for later in my opinion.

Nytelife26:
It occurs to me that actually, there is a more sensible way to do this. Many of our checks are formatted like so: `check.subcheck`. There is never any `check.subcheck.subsubcheck`. Why, then, are we grouping several subchecks into one file?

An easy way to solve this would be to make them individual, as they should be.

PR #985, opened 2019-08-06
Fix incorrectly placed mondegreen

denizgenc:
The file labelled `"attorney and notary public"` as a mondegreen, and proposes `"attorney and not a republic"` as the preferred form, which was incorrect. This change swaps them around to fix it.


suchow:
Good eye, thanks for the contribution @denizgenc and thanks for the review @Nytelife26.

PR #1008, opened 2019-11-04
Test on Python 3.7 and 3.8

polyzen:


polyzen:
Seems you deactivated your Travis account?

PR #1065, opened 2020-04-16
fix: correct error message of annotations.misc

howsiwei:


Nytelife26:
Thank you for the PR. Given that categorization is an ongoing effort, your work is appreciated.

PR #1081, opened 2020-06-10
feat: add --config cli argument

### LINKED ISSUE #356: Config files are loaded only from a central location
The `.proselintrc` file is [searched for](https://github.com/amperser/proselint/blob/master/proselint/command_line.py#L56) only in the base directory of proselinty itself.  This makes it impossible for the user to disable some of the checking rules.

### LINKED ISSUE #1091: Allow disabling individual checks

reiddraper:
This is useful for projects that want to provide their own proselintrc
which is stored in the repository for the project itself. Further, it's
useful for installations where putting a config-file outside of a
sandboxed directory is not possible.

---

Resolves #356. Resolves #1091.


wargio:
can we have this merged asap? i do not see why this wasn't already an option in the tool.

carlsmedstad:
Apart from needing a rebase on `main` and some clean up to pass the CI, this looks good to me.

Should @reiddraper continue this work? @Nytelife26 @suchow

Nytelife26:
Thanks for bringing this to my attention, @carlsmedstad. I'll get to work on bringing it up to date and merging now, since it is a useful feature. We could also put automatically detecting proselintrc files up to the filesystem boundary as part of this PR, too, but that's another thing.

carlsmedstad:
Great, detecting project specific config files seems like something that would be useful after this has been merged.

I'm a proponent of the pattern of having three levels of configuration: system, user and project scope.

Nytelife26:
> I'm a proponent of the pattern of having three levels of configuration:
> system, user and project scope.

Me too ultimately. The user configuration should check `~/.config` too, though.

Nytelife26 (approved):

carlsmedstad:
Definitely, I could make a PR tomorrow if you'd like?

suchow:
@Nytelife26 The CI run is failing here for one substantive reason (coverage) and one transition-to-Github-actions reason. In particular, something is breaking during the Danger action.

Nytelife26:
> @Nytelife26 The CI run is failing here for one substantive reason
> (coverage) and one transition-to-Github-actions reason. In particular,
> something is breaking during the Danger action.

I am aware of the coverage issue. The test suite is not currently very
comprehensive, it seems.

As for the Danger action, I believe that is a result of the force push I
performed following the rebase, given that Danger specifies that as a
potential cause in the error log.

Nytelife26 (approved):

Nytelife26:
cc: @suchow

tests/test_config_flag.py
```diff
@@ -0,0 +1,16 @@
+"""Check that the --config command-line flag works"""
```
suchow, line 1: The test name and docstring suggest that this is testing the `--config` option, but the body is testing the `load_options` functions, which is one part of the pipeline that underlies the `--config` option handling. Did you test the end-to-end pipeline using the CLI tool? Minimally you could confirm you tested it manually. More fully, you could include a test of the `proselint()` function (setting the `config` kwarg) or even the `proselint` CLI command with the `--config` option set and not set.

suchow: Also, we're not using `--config` as a "flag" here, but as an "option".

suchow (changes requested):

suchow (commented):
The new test looks good, just need to `import subprocess` up top and I think it should be good to go!

benknoble:
This doesn't resolve #1091, which is about individual checks versus categories.

PR #1134, opened 2021-04-22
chore: modernize proselint

Nytelife26:
So far:

- fix: update clock for latest python
- chore: bootstrap poetry
- chore: remove requirements in favour of poetry
- chore: update dependencies
- test: fix test suite
- fix: add warning for broken corpus
- chore: remove deprecated setup lines
- docs: update contribution guidelines

Upcoming:

- [x] Add GH workflows for CI (instead of Travis and AppVeyor)
  - [x] All Python versions 3.6-3.9
    - [x] Unix support
    - [x] Windows support
  - [x] Linting
  - [x] Testing (`pytest` and `coveralls`, dropping `nose`)


Nytelife26:
cc: @suchow (not yet ready for review, just tagging you so you can inspect work so far if you want and we can discuss here)

proselint/command_line.py
```diff
@@ -27,6 +27,7 @@
 demo_file = os.path.join(proselint_path, "demo.md")
 
 
+# TODO: fix broken corpus
```
suchow, line 30: FYI, the corpus can be found at https://github.com/amperser/corpus, but because we don't have the rights to distribute the texts, it's a private repo. I'll give you access soon. If you can think of a workaround that allows us to keep the test, let me know.

tests/test_exit_codes.py
```diff
@@ -6,7 +6,7 @@
 def test_exit_code_demo():
     """Ensure that linting the demo returns an exit code of 1."""
     try:
-        subprocess.check_output(["proselint", "--demo"])
+        subprocess.check_output(["python", "-m", "proselint", "--demo"])
```
suchow, line 9: I'm familiar with the `-m` flag, but can you say something about what's changed that requires proselint to be run in this way?

Nytelife26: It isn't a requirement, so much of a better way to run tests in a sandboxed environment. For development testing, people often may want to keep the package sandboxed, and therefore not install it to their global `site-packages` - this allows them to do that and still run the tests successfully.

proselint/command_line.py
Nytelife26: Interesting. I cannot think of a workaround at the moment, however, it is an intriguing feature. I must ask, though, what is the point in including the option in the end software if it is broken by default without the corpus and people cannot obtain it?

suchow: There’s no reason to include it in the end software, could splinter it off into a developer only tool or have it appear only in tests

Nytelife26: That'd make more sense. In that case, I'll remove the option altogether for now.

suchow: Great, thanks. Can you open an issue for it?

suchow:
Just wanted to say thanks for your work on this, it’s looking great so far! It’s exciting to see the gears beginning to turn again.

Nytelife26:
> Just wanted to say thanks for your work on this, it’s looking great so far! It’s exciting to see the gears beginning to turn again.

I'm happy to work on it. Honestly, it's an awesome concept, and I've already shown it to some of my dyslexic friends. They're very pleased by the idea of keeping writing concise and avoiding generic cliches.

I need the ability to test the new workflows, though, so I'll see if I can make that work on my fork.

Nytelife26:
Beautiful. [Over 16 test runs across all supported versions and systems, and linting in under 4 minutes](https://github.com/Nytelife26/proselint/actions/runs/803410065).

suchow:
Hurray!

Nytelife26:
I think aside from any additional CI, perhaps PR labels and mergify, this is pretty much ready. Anything else in scope here, @suchow?

suchow:
@Nytelife26 I think this is plenty. My understanding is that this PR replaces the existing CI tooling (Travis, AppVeyor), so the last uncertainty I have is about how to transition from one to the other. Should we just merge it and then fix any issues that arise? In a higher-stakes environment that would seem unwise, but in our case I don't see a problem.

Nytelife26:
@suchow Unfortunately there aren't really any ways to test most of these changes beyond making the fork and testing them there itself, so we may have to. In a higher stakes environment I'd be using an enterprise CI like ConcourseCI, but that's more of a corporate with maintainable bus factor kind of deal. Fortunately the worst that can happen here since the only CI I couldn't test were automated releases and website updating is that it fails to update the website or fails to release to PyPI

suchow:
@Nytelife26 Okay, well I'll wait until you're ready to merge, but it looks good to me.

Nytelife26:
> @Nytelife26 Okay, well I'll wait until you're ready to merge, but it
> looks good to me.

I'm pretty confident this is a good first step to getting this show back
on the road. I reckon we should put this on the road and test it out \-
after all, worst case scenario, we can just fix it.

PR #1147, opened 2021-05-16
Fix lint errors on master

suchow:


suchow:
@Nytelife26 Do you know what's going on here? I merged #1134 into master and all the CI checks passed except for the linter, so I created the present PR that fixes the code to run cleanly using `pydocstye` and `pycodestyle`. However, when the CI runs over this PR, it fails with the following error:

```
Run poetry run coverage run -m pytest --continue-on-collection-errors
/home/runner/work/_temp/e623ed8c-c5bd-49e0-b795-71b524791775.sh: line 1: poetry: command not found
Error: Process completed with exit code 127.
```
which doesn't make much sense to me given that this PR doesn't touch any of the CI-relevant code.

Nytelife26:
Apologies, I was asleep. And that's certainly interesting. I'll investigate now after I accept the invitation to the repository.

Nytelife26:
**UPDATE**: Seems largely like a problem with dependency caching - evidently, restoring the pip cache alone does not seem to successfully restore poetry, like it would've with npm. It may be possible that after restoring the cache I'll have to make the CI run an extra pip command to initialize it properly, or drop caching altogether.

Nytelife26:
Unfortunately i cannot yet test my fixes due to GitHub Actions being down (they have been queued for over 10 minutes). I will re-run the jobs later when this is resolved.

image

PR #1150, opened 2021-05-16
fix: correct always-passing uncomparables

Nytelife26:
- ci: remove fail-fast test matrix
- test: use concatenation instead of append and 1st index
- fix: patch always-passing uncomparables
- ci: remove pypy from testing

### DIFF
--- a/proselint/checks/uncomparables/misc.py
+++ b/proselint/checks/uncomparables/misc.py
@@ -45,11 +45,9 @@ One axiom of Standard Written English is that your reader is paying close
 attention and expects you to have done the same.
 """
 import re
-from proselint.tools import memoize
 import itertools
 
 
-@memoize
 def check(text):
     """Check the text."""
     err = "uncomparables.misc"
@@ -114,7 +112,7 @@ def check(text):
         ("more", "possible")  # FIXME
     ]
 
-    all = [r"\\b" + i[0] + r"\s" + i[1] + r"[\W$]" for i in itertools.product(
+    all = [i[0] + r"\s" + i[1] + r"[\W$]" for i in itertools.product(
            comparators, uncomparables) if i not in exceptions]
 
     occ = re.finditer("|".join(all), text.lower())
--- a/tests/check.py
+++ b/tests/check.py
@@ -32,9 +32,9 @@ class Check(TestCase):
 
         errors = []
         for text in lst:
-            errors.append(self.this_check.check(text))
+            errors += self.this_check.check(text)
 
-        return len(errors[0]) == 0
+        return len(errors) == 0
 
     def wpe_too_high(self):
         """Check whether the check is too noisy."""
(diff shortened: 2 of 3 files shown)


tests/check.py
```diff
@@ -32,9 +32,9 @@ def passes(self, lst):
 
         errors = []
         for text in lst:
-            errors.append(self.this_check.check(text))
+            errors += self.this_check.check(text)
 
-        return len(errors[0]) == 0
+        return len(errors) == 0
```
suchow, line 37: This isn't right — checks returns errors that are lists with the error code, location, message, etc. `append` adds each of these elements to the list, whereas we want the entire list added to the list. In this case, it makes no difference because we're only testing whether the list of errors is empty (and so your switch to `+=` doesn't break anything), but in general it's good to keep in mind that errors are returned as lists.

proselint/checks/uncomparables/misc.py
```diff
@@ -45,11 +45,9 @@
 attention and expects you to have done the same.
 """
 import re
-from proselint.tools import memoize
```
suchow, line 48: If this fixed anything, it's because the results of previous test runs were getting cached and so even once you fixed the problem you were still seeing it fail. I saw the same thing. The fix isn't to remove memoization here, it's to clear the cache/shelf between test runs, which must not be happening correctly.

suchow:
I incorporate the error you found via https://github.com/amperser/proselint/commit/8a1cdc1492bf642196397caec9fe56e9fb521eb8, the other changes you made here I believe may have been side effects of hunting for the real bug. For example, I've opened https://github.com/amperser/proselint/issues/1151 to deal with the issue of memoization across test runs.

proselint/checks/uncomparables/misc.py
Nytelife26: You are absolutely correct, that's something I forgot to re-implement after I took it out during my testing.

tests/check.py
Nytelife26: The thing I foresee with this is that it can potentially call `append` multiple times depending on the number of checks that are passed. If it only checks the length of the first list it adds, what's to say the rest don't contain errors?

suchow: Oops, you're right. Sorry about that. Please excuse me if it seems like I don't know anything about Proselint, it's been a while ;)

Can you open a new PR with just the changes on lines 35/37?

Looks like this must have always been broken. Maybe none of the tests ever send in a list of sentences vs. one at a time…

Nytelife26: That's okay, honestly I'm just grateful that you're attentive and sharing your thoughts on things. Sure, though, I'll open a new PR shortly.

Although, it does seem that no tests ever send in a list. Perhaps it would make testing faster to convert some of them, given that a loop inside of a function would be significantly faster than calling the function several times.

PR #1152, opened 2021-05-17
fix: reduce false positives in uncomparables

Nytelife26:
Prior to this, something like "Every perfect instance" would flag as "very
perfect" under the uncomparables check. This PR fixes that.


suchow:
Looks good, why don’t you add a test using the example you give in the comment and then we’ll merge it?

Nytelife26:
Good idea, not entirely sure why I didn't think of that.

PR #1170, opened 2021-05-28
test: add curly quotes test for typography

Nytelife26:
## This relates to...

The symbols typography check, specifically the curly quotes subcheck.
Prior to this PR it raised inapplicable false positives when matching
lone quotes instead of checking for pairs, and the test function for
it was missing.

## Changes

- fix(checks/typography): adjust curly quotes regex
- test: add curly quotes typography test

### Features

N/A.

### Bug Fixes

- Fixed a faulty matcher for the curly quotes typography check that
  would match any quotes instead of a pair.

N/A.


suchow (changes requested):
The description of the PR in this discussion doesn't match the new tests. In particular, the description says that the reason for the adjustment is to avoid false positives with lone quotes, but I see no tests in which there is a lone quote.

suchow (approved):

PR #1172, opened 2021-06-01
make checks individual

### LINKED ISSUE #491: .proselintrc usage

### LINKED ISSUE #729: Support more than two keys in .proselintrc

### LINKED ISSUE #1091: Allow disabling individual checks

Nytelife26:
## This relates to...

The structure of checks directories.

**Related issues:**

- Follows on from #951.
- This resolves #729.
- This resolves #491.
- This resolves #1091.

## Rationale

At present, some subcheck files (e.g. `typography.symbols`, `redundancy.misc`)
contain multiple subchecks. This is an illogical structure, prevents disabling
the subchecks in `.proselintrc` which should be possible, et cetera. This PR
fixes that by moving them into individual subcheck files, where they belong.

## Changes

N/A (WIP).

### Features

N/A.

### Bug Fixes

N/A.

N/A.


Nytelife26:
As an update, I think it would be better to make a plan for recategorising things before actually implementing the changes. Ideally, there will be some kind of framework as to what kind of checks are supported and determining where they belong.

PR #1174, opened 2021-06-07
use regex for lexical illusions

Nytelife26:
## This relates to...

The usage of regular expressions rather than simple matches for
`lexical_illusions.misc`.

## Rationale

Prior to this, the lexical illusions check would only catch
3 different lexical illusions, and always in pairs of two rather than
flagging the entire strand. This is inefficient, and does not catch
many lexical illusions at all.

## Changes

- `lexical_illusions.misc` now uses regex
- `lexical_illusions.misc` now reports entire strands instead of many word pairs
- `lexical_illusions.misc` reports all lexical illusions except for `that that`
  and `had had`

### Features

N/A.

### Bug Fixes

N/A.

N/A.

### DIFF
--- a/proselint/checks/lexical_illusions/misc.py
+++ b/proselint/checks/lexical_illusions/misc.py
@@ -21,11 +21,8 @@ def check(text):
     """Check the text."""
     err = "lexical_illusions.misc"
     msg = u"There's a lexical illusion here: a word is repeated."
+    regex = r"\b(\w+)(\b\s\1)+"
+    exceptions = [r"^had had$", r"^that that$"]
 
-    list = [
-        r"the\sthe",
-        r"am\sam",
-        r"has\shas"
-    ]
-
-    return existence_check(text, list, err, msg)
+    return existence_check(text, [regex], err, msg, exceptions=exceptions,
+                           require_padding=False)
--- a/tests/test_lexical_illusions.py
+++ b/tests/test_lexical_illusions.py
@@ -20,3 +20,4 @@ class TestCheck(Check):
         """Basic smoke test for lexical_illusions.misc."""
         assert self.passes("""Smoke phrase with nothing flagged.""")
         assert not self.passes("""Paris in the the springtime.""")
+        assert self.passes("""And he's gone, gone with the breeze""")
--- a/tests/test_lexical_illusions.py
+++ b/tests/test_lexical_illusions.py
@@ -18,6 +18,9 @@ class TestCheck(Check):
 
     def test_smoke(self):
         """Basic smoke test for lexical_illusions.misc."""
-        assert self.passes("""Smoke phrase with nothing flagged.""")
-        assert not self.passes("""Paris in the the springtime.""")
-        assert self.passes("""And he's gone, gone with the breeze""")
+        assert self.passes("Smoke phrase with nothing flagged.")
+        assert not self.passes("Paris in the the springtime.")
+        assert self.passes("And he's gone, gone with the breeze")
+        assert self.passes("You should know that that sentence wasn't wrong.")
+        assert self.passes("She had had dessert on the balcony.")
+        assert not self.passes("You should know that that that was wrong.")


proselint/checks/lexical_illusions/misc.py
```diff
@@ -21,11 +21,6 @@ def check(text):
     """Check the text."""
     err = "lexical_illusions.misc"
     msg = u"There's a lexical illusion here: a word is repeated."
+    regex = r"\b(\w+)\b\s\1"
```
suchow, line 24: There are a few instances where repeated words are perfectly acceptable. For example, one day I was commenting on student writing and I noticed that the student used the word "that" instead of "which". It's a hard distinction for many, and I don't blame them. So here I am now, telling you that that "that" that that student wrote ought to have been a "which".

suchow: See also https://en.wikipedia.org/wiki/Buffalo_buffalo_Buffalo_buffalo_buffalo_buffalo_Buffalo_buffalo

suchow: https://en.wikipedia.org/wiki/James_while_John_had_had_had_had_had_had_had_had_had_had_had_a_better_effect_on_the_teacher

Nytelife26:
> There are a few instances where repeated words are perfectly acceptable.

Gramatically valid, perhaps. Stylistically speaking, it's an absolute
atrocity. I had to attempt to read that example you threw in around 3
times before it parsed correctly in my mind.

Valid, yes - but also beneficial (and always possible) to avoid. An
exception has been made for punctuation inbetween word boundaries,
because splitting the lexical structure into phrases makes it readable.

I can therefore tell you that the "that" that the student wrote should
have been a "which", but there are other ways of writing about the
"which" which the student should've written without entering linguistic
territory in which a "which" which your student should've written such
that the "that" that your student wrote isn't so bad.

Nytelife26:
> https://en.wikipedia.org/wiki/James_while_John_had_had_had_had_had_had_had_had_had_had_had_a_better_effect_on_the_teacher

"is an English sentence used to demonstrate lexical ambiguity and the
necessity of punctuation"

I do not think from a stylistic perspective we should be referencing
esoteric sentences that serve to demonstrate poor linguistic usage when
justifying limitations.

```text
The sentence is easier to understand with added punctuation and
emphasis:

    James, while John had had "had", had had "had had"; "had had" had
    had a better effect on the teacher.
```

The tool is designed to help people write better. Avoiding structures
like the ones you have referenced in favour of their better formatted
alternatives is doing exactly that, I feel.

Either way, the decision lies with you. I've said my piece.

suchow:
How about we make an exception for `that that` and for `had had` (which is the past perfect form of the verb to have and hardly a verbal atrocity, e.g., "He had had one too many drinks."), and you can have the rest? The examples were meant to be fun examples, not things you needed to take seriously, which is why I left them as comments and didn't mark the changes as necessary in the review.

Nytelife26:
> How about we make an exception for `that that` (two in a row, not more) and for `had had` (which is the past perfect form of the verb to have and hardly a verbal atrocity, e.g., "He had had one too many drinks."), and you can have the rest?

Works for me. Apologies if that came off aggressive, by the way, I was just sharing my thoughts and it ended up coming out somewhat passionately

suchow:
All good, not going to complain about passion!

Nytelife26:
It seems like to do so we will need to be able to create exceptions using the
existence check. Unfortunately, Regex does not present us with a viable way to
do this (no amount of negative lookaround hacks would work), so we may need to
use a separate PR to implement that - quite a breaking change.

Nytelife26:
@suchow The exceptions for `had had` and `that that` have been successfully implemented. Furthermore, the lexical illusions check will now flag the entire strand of lexical illusions, rather than repeatedly flagging it in word pairs of two - much more efficient.

suchow (approved):
Wonderful, thank you, an excellent way to resolve this issue.

Issue #1151, opened 2021-05-17
Clear cache between pytest runs


suchow:
When trying to fix an issue with the uncomparables check, @Nytelife26 removed the memoization on that function, presumably  because the test suite was not passing even though the function was fixed. A better solution is for us to figure out how to get pytest to clear the cache/shelf between runs.

Here's the relevant memoization code:

https://github.com/amperser/proselint/blob/8a1cdc1492bf642196397caec9fe56e9fb521eb8/proselint/tools.py#L99

Nytelife26:
This seems to be a culmination of various issues. I suspect I'll need to
take a further look through how the shelving system and the cache
dropping system works. I found an error not long ago with the
misplacement of an unbound variable in the memoize function so perhaps
there's more to this than it seems.

Nytelife26:
I believe the cache system may need a redesign, or at the very least the package *should* clear the cache upon installation to avoid issues persisting between versions.

joshmgrant:
If you need help with this task, I can pick it up or help. I wrote most of the original pytest tests.

Nytelife26:
> If you need help with this task, I can pick it up or help. I wrote most
> of the original pytest tests.

I'd just like to say I greatly appreciate your volunteering and it's
awesome to see some of the original engineers still hanging around. You
can if you'd like to, any contributions are appreciated :)

joshmgrant:
Just so I have the right context: was the issue with cache interfering with tests on local developer runs (eg on a laptop when someone is developing) or on GitHub Actions or both? I can add more logic if it's the latter case.

Nytelife26:
GitHub Actions are fine since they are run in an individual environment each time, @joshmgrant. The issue lies with local environments on development machines since they do not clear the cache, and the user has to clear it manually.

The best solution to this is to reset the cache at the path on `~/.cache/proselint` in the installation script for proselint and also for test runs. Perhaps the test suite could disable memoization or use a cache bypass? Unsure.

Nytelife26:
@joshmgrant was able to coordinate a solution by bypassing the cache altogether in test runs. This will be resolved soon.

PR #1176, opened 2021-06-07
test: bypass cache in pytest runs

joshmgrant:
Starting a PR to address #1151

### DIFF
--- a/tests/check.py
+++ b/tests/check.py
@@ -31,7 +31,7 @@ class Check(TestCase):
 
         errors = []
         for text in lst:
-            errors.append(self.this_check.check(text))
+            errors.append(self.this_check.check.__wrapped__(text))
 
         return len(errors[0]) == 0
 
@@ -51,7 +51,7 @@ class Check(TestCase):
             # Compute the number of words per (wpe) error.
             with codecs.open(example_path, "r", encoding='utf-8') as f:
                 text = f.read()
-                num_errors = len(self.this_check.check(text))
+                num_errors = len(self.this_check.check.__wrapped__(text))
                 num_words = len(text)
 
             try:
--- a/tests/check.py
+++ b/tests/check.py
@@ -17,8 +17,7 @@ class Check(TestCase):
 
     def tearDown(self):
         """Create a placeholder for teardown procedure."""
-        from proselint.tools import close_cache_shelves
-        close_cache_shelves()
+        pass
 
     @property
     def this_check(self):


Nytelife26:
Thank you for the pull request! This is, at present, a much-needed fix.

However, Pytest's cache is not the cause of the issue, unfortunately. All Pytest does is cache test file IDs to make collection faster. The real issue lies within the caching used by our lint functions with the memoize decorator - this is not cleared between tests currently, which causes runs to be oblivious to source code changes.

joshmgrant:
Thanks for the feedback! I feel like I remember this issue (or something like that) from a long time ago. I'll see if I can rework some of the `test\` code to help with this issue.

Nytelife26:
@joshmgrant That doesn't remove the shelving, it just takes out the part where it gets closed afterwards.

Nytelife26:
I'll push what I've been working on. At present, I'm making a function called
in the setup for each test that **should** clear the cache for the check.
Emphasis on should. At the minute, something seems to be going wrong with
tests, though, so that needs to be fixed.

If you'd like, you can take a look or pick up where I left off. For now, I'm
signing off for the day, so I'll continue tomorrow.

joshmgrant:
Yeah, I was hoping I could use a quick hack to get around the caching issue. It looks like my idea didn't work :)

I guess the main question I have is: should we find a way to test the individual checks without memoization? Or should we find a better way to structure the pytest tests to allow for caching somehow?

Nytelife26:
> I guess the main question I have is: should we find a way to test the
> individual checks without memoization? Or should we find a better way to
> structure the pytest tests to allow for caching somehow?

It would be ideal to bypass the cache lookup or clear the cache prior to
the tests. The former would be better, but I believe it would also be
significantly more complicated to implement - unless we use an
environment variable that gets set in the setup of the test suite, and
then in the memoization function if the variable is present it will not
index the cache.

joshmgrant:
So I was able to find a way to get undecorated functions imported into a module. This means we can test the "unmemoized" methods directly from each check module/class in the tests. The only catch is - wait for it - this currently only works for Python 3. I can see if I can update the logic for Python 2 and Python 3 if needed. Let me know if this is what you're looking for!

Nytelife26:
That's certainly something. Honestly I didn't know you could do that, so it's certainly a learning experience. Fortunately, that's not a catch - we recently decided to drop support for Python 2 anyway.

[Nytelife26 - Chat @ Spike](https://spikenow.com/r/a/?ref=spike-organic-signature&_ts=132n80)	[132n80]

On June 22, 2021 at 3:38 GMT, joshin4colours ***@***.***> wrote:

So I was able to find a way to get undecorated functions imported into a module. This means we can test the "unmemoized" methods directly from each check module/class in the tests. The only catch is - wait for it - this currently only works for Python 3. I can see if I can update the logic for Python 2 and Python 3 if needed. Let me know if this is what you're looking for!

Nytelife26:
@joshmgrant I altered your latest commit. Rather than changing the `this_check.check` to `this_check(...)` and then altering it to return the wrapped version in every file, we can simply use `this_check.check.__wrapped__` since there will never be a not-memoized check and there will never be a test run that doesn't want the original check.

Nytelife26:
I can confirm, these changes have fixed the problem. Brilliant work, @joshmgrant.

I'll experiment with some better ways of doing this as I mentioned in my previous comment, and then we can get this reviewed and merged.

Nytelife26:
@suchow This is ready for review if you wouldn't mind :)

Nytelife26 (approved):

PR #1183, opened 2021-07-04
chore: fix package metadata

Nytelife26:
## This relates to...

The package metadata and installation setup.

## Rationale

Prior to this, the package wheel would not build properly, and there
were erroneous setup files from the old package management system.

## Changes

- removed `setup.py` and `setup.cfg`
- added metadata to `pyproject.toml`
- added `.bumpversion.cfg`
- reformatted `README.md` and fixed some phrasing


Nytelife26:
Unfortunately, we cannot fix the Danger issue until we merge this and cut a new release.

suchow (changes requested):
Looks good, my only complaint is hard wrapping the Markdown, which isn’t necessary given that most editors can soft wrap Markdown text and keeping paragraphs on one line makes diffs more readable.

Nytelife26:
> Looks good, my only complaint is hard wrapping the Markdown, which
> isn’t necessary given that most editors can soft wrap Markdown text
> and keeping paragraphs on one line makes diffs more readable.

That's something my editor does automatically, apologies - let me go fix
that quickly.

suchow (approved):

PR #1188, opened 2021-07-05
feat: convert meta-checks to decorators

Nytelife26:
## This relates to...

Adding threshold checks. This blocks #802. This is blocked by #1217 and #1218.

## Rationale

This is useful for situations such as in
`checks.typography.exclamation.30ppm` where a PPM threshold metric
determines whether to raise an error or not.

## Changes

- Add `tools.threshold_check`
- Convert `checks.typography.exclamation` to use `threshold_check`

### Features

- Add `tools.threshold_check`

### Bug Fixes

N/A.

N/A.

### DIFF
--- a/proselint/checks/typography/exclamation.py
+++ b/proselint/checks/typography/exclamation.py
@@ -12,9 +12,7 @@ categories: writing
 Too much yelling.
 
 """
-import re
-
-from proselint.tools import existence_check, max_errors, memoize
+from proselint.tools import existence_check, max_errors, memoize, ppm_threshold
 
 
 @max_errors(1)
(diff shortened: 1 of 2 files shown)


suchow:
Rather than implementing this as a meta-check, I think we should either:

1. Create a decorator `@threshold` that takes a parameter specifying the threshold ppm below above which errors will be flagged.
2. Do what we are currently doing with `max_errors` via the `truncate_to_max` function.

I think I'd put in a vote for making both `threshold` and `max_errors` decorators that would work for all check types.

suchow (dismissed):
Rather than implementing this as a meta-check, I think we should either:

1. Create a decorator `@threshold` that takes a parameter specifying the threshold ppm below above which errors will be flagged.
2. Do what we are currently doing with `max_errors` via the `truncate_to_max` function.

I think I'd put in a vote for making both `threshold` and `max_errors` decorators that would work for all check types.

Issue #1191, opened 2021-07-06
False alarm on lexical illusions


suchow:
Test document:

```
the theory of relativity
```

Output:

```
(base) ➜  Downloads proselint issue.md
issue.md:1:2: lexical_illusions.misc There's a lexical illusion here: a word is repeated.
```

[issue.md](https://github.com/amperser/proselint/files/6766624/issue.md)

suchow:
Here are a bunch of examples:

- `The theory` (the the)
- `the practitioner's side` (s s)
- `an angle` (an an)
- `Applying Newton's second law` (s s)
- `an analytical solution` (an an)

PR #1192, opened 2021-07-06
fix: correct false positives for lexical illusion regex

suchow:
Hotfix for #1191


suchow:
Screen Shot 2021-07-05 at 10 22 25 PM

Screen Shot 2021-07-05 at 10 22 32 PM

suchow:
@Nytelife26 If this looks okay to you, let's cut a release asap.

PR #1212, opened 2021-08-02
fix: correct config export

### LINKED ISSUE #1200: Problems packaging .proselintrc
Hey :wave:

I'm running into issues packaging this for Arch due to the system config file `.proselintrc`. Since the project migrated from `setuptools` to `poetry` I use [dephell](https://github.com/dephell/dephell) to convert it back to `setuptools` during installtion, as is recommended  by Arch community. It does seem like this conversion does not produce the correct `package_data` in `setup.py` (probably due to this not being supported) resulting in `.proselintrc` not being installed at all.

While I could patch the package and do a number of work-arounds, I wonder if we could get rid of `.proselintrc` altogether by hardcoding the system defaults directly in `proselint`. Having `.proselintrc` as a Python file or a part of `proselint/tools.py` would make packaging for distros less complex and it would not hurt `pip` distribution as far as I can tell.

Not distributing the default config in file format would make it a bit harder to make a custom config from scratch, but this could be rectified by a `--dump-config` flag which would print the current config. This is how the Clang tools works, for example.

Nytelife26:
## This relates to...

The packaging of our default `.proselintrc`.

This resolves #1200.

## Changes

- Added `--dump-config` and `--dump-default-config`
- `proselint.tools.lint` now takes a config object instead of using
  `load_options`
- Now exports `proselint.config.default`

### Features

- Added `--dump-config` and `--dump-default-config`
- Now exports `proselint.config.default`

### Bug Fixes

- `proselint.config.default` is now packaged alongside `proselint`, fixing
  packaging for distros.

- `proselint.tools.lint` now takes a config object instead of using
  `load_options`


proselint/config.py
```diff
@@ -0,0 +1,87 @@
...
+        "misc.capitalization": True,
+        "misc.chatspeak": True,
+        "misc.commercialese": True,
+        "misc.composition": True,
+        "misc.currency": True,
+        "misc.debased": True,
+        "misc.false_plurals": True,
+        "misc.illogic": True,
+        "misc.inferior_superior": True,
+        "misc.institution_name"	: True,
```
suchow, line 40: Some whitespace issue here.

proselint/tools.py
```diff
@@ -192,10 +184,12 @@ def load_options(config_file_path=None):
         if os.path.isfile(path):
             user_options = json.load(open(path))
             break
+        if os.path.isfile(path.replace(".json", "")):
```
suchow, line 187: I might not be following what this is doing. Will this raise an exception if someone has a .proselintrc file but no .proselintrc.json file? It so, it might be better to give a deprecation warning using `warnings.warn` (https://docs.python.org/3/library/warnings.html#warnings.warn) and still process the .proselintrc file.

proselint/command_line.py
```diff
@@ -95,15 +97,25 @@ def print_errors(filename, errors, output_json=False, compact=False):
...
+def proselint(paths=None, config=None, version=None, clean=None,
+              debug=None, output_json=None, time=None, demo=None, compact=None,
+              dump_config=None, dump_default_config=None):
     """Create the CLI for proselint, a linter for prose."""
+    if dump_default_config:
+        return print(json.dumps(default, sort_keys=True, indent=4))
+
+    config = load_options(config, default)
+    if dump_config:
+        return print(json.dumps(config, sort_keys=True, indent=4))
```
suchow, line 114: There's something mildly grating about returning the `None` output of `print`. Not a big deal, and I can't really explain the feeling, but I prefer the original.

suchow (approved):

proselint/command_line.py
Nytelife26: I get you, honestly. It's just a shorter more concise way to return - but, you're right, perhaps we shouldn't return from a function call that evaluates to a void expression and instead just use `return` for clarity.

proselint/tools.py
Nytelife26: Honestly, I wasn't aware of the existence of a `warnings` library - I have yet to deal with a package deprecation in Python, largely. Always fun to learn new things about a language even over 7 years after starting to use it :)

PR #1215, opened 2021-08-11
Switch instances of master to main in README and CONTRIBUTING.

tdenewiler:


Nytelife26:
You're right - we're working on ways to fix that for external PRs. It's difficult with the way GitHub's API permissions work.

PR #1246, opened 2021-11-16
fix: load default config if none specified

### LINKED ISSUE #1240: KeyError in Python
I installed proselint with pip. Running the example in the readme:
 
```py
import proselint
suggestions = proselint.tools.lint("This sentence is very unique")
```
gave me this output:

```
Traceback (most recent call last):
  File [path_removed], line 2, in <module>
    suggestions = proselint.tools.lint("This sentence is very unique")
  File "[path_removed]/venv/lib/python3.9/site-packages/proselint/tools.py", line 240, in lint
    checks = get_checks(config)
  File "[path_removed]/venv/lib/python3.9/site-packages/proselint/tools.py", line 142, in get_checks
    check_names = [key for (key, val) in options["checks"].items() if val]
KeyError: 'checks'
```

[The reason for this appears to be because the "lint" function makes "config" an empty dict by default](https://github.com/amperser/proselint/blob/f7faa720ea72cba3b0f7aa552fbf4dc62478e170/proselint/tools.py#L233). Perhaps having it load the default config instead would be a good solution?

I used Python 3.9.6 on MacOS 10.15.4 (Catalina). This is my first time submitting an issue, so let me know if I've done anything wrong.

oxytocin:
Avoids crash in Python if configuration is not specified. Fixes #1240.

### DIFF
--- a/proselint/tools.py
+++ b/proselint/tools.py
@@ -13,6 +13,7 @@ import shelve
 import sys
 import traceback
 from warnings import showwarning as warn
+
 from . import config
 
 _cache_shelves = dict()
--- a/proselint/tools.py
+++ b/proselint/tools.py
@@ -13,6 +13,7 @@ import shelve
 import sys
 import traceback
 from warnings import showwarning as warn
+from . import config
 
 _cache_shelves = dict()
 proselint_path = os.path.dirname(os.path.realpath(__file__))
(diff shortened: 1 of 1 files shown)


proselint/tools.py
```diff
@@ -13,6 +13,7 @@
 import sys
 import traceback
 from warnings import showwarning as warn
+from . import config
```
Nytelife26, line 16:
```suggestion

from . import config
```
Separating local or external imports from builtins are necessary in lint rules.

Nytelife26 (changes requested):

Nytelife26 (approved):
LGTM, thank you for the PR :)

PR #1295, opened 2022-04-29
docs: add neovim plugin with null-ls

darcyparker:


Nytelife26:
Interesting, I had no idea null-ls supported proselint! I'll use the plugin myself for a short while to test it and then approve if it works well.

PR #1315, opened 2022-09-11
Add support for disabling checks via comments

Holzhaus:
In some cases, there is a false positive or the writer knowingly wants
to ignore a failing check *in a specific section of the text*, but not
the entire document.

With this change, checks can be enabled and re-enabled at any time using
special comments:

    Here, all checks are used.
    % proselint: disable
    No checks are enabled here.
    % proselint: enable
    All checks have been reenabled.

It's also possible to disable only some checks:

    Here, all checks are used.
    % proselint: disable=nonwords.misc,weasel_words.very
    Here, the \texttt{nonwords.misc} and \texttt{weasel\_words.very}
    checks are disabled.
    % proselint: enable=nonwords.misc
    At this point, the \texttt{nonwords.misc} check has been reenabled.

This aims to be markup-language agnostic by allowing any printable
non-word ASCII character  in front of `proselint: instead of maintaining
a list of allowed comment styles. Note that there mustn't be any word
characters, so toggling checks in the middle of a line is not possible,

### DIFF
--- a/proselint/tools.py
+++ b/proselint/tools.py
@@ -248,10 +248,11 @@ def lint(input_file, debug=False, config=config.default):
 
         for error in result:
             (start, end, check, message, replacements) = error
+            if is_quoted(start, text):
+                continue
             (line, column) = line_and_column(text, start)
-            if not is_quoted(start, text):
-                errors += [(check, message, line, column, start, end,
-                            end - start, "warning", replacements)]
+            errors += [(check, message, line, column, start, end,
+                        end - start, "warning", replacements)]
 
         if len(errors) > config["max_errors"]:
             break
--- a/proselint/tools.py
+++ b/proselint/tools.py
@@ -1,6 +1,7 @@
 """General-purpose tools shared across linting checks."""
 
 import copy
+import collections
 import dbm
 import functools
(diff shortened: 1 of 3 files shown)


Holzhaus:
Ping, this would come in handy at times.

proselint/tools.py
```diff
@@ -230,6 +231,48 @@ def line_and_column(text, position):
     return (line_no, position - position_counter)
 
 
+def find_ignored_checks(available_checks, text):
+    ignored_checks = collections.defaultdict(dict)
+    for match in re.finditer(r"^[!-/:-@[-`{-~\s]*proselint: (?P<action>disable|enable)(?P<checks>(?:=(?:[\w\.,]*)))?", text, flags=re.MULTILINE):
```
Nytelife26, line 236: Consider the computational cost of this. RegExes are expensive and moderately slow, so running one on every line just to determine whether or not we can disable some checks feels inefficient.

I am not quite sure what the solution to this would be yet, but in the meantime any ideas are welcome. They may not be quite as simple or elegant, but proselint's performance is lacking as it stands, and implementing this would be a major blow in the wrong direction.

Issue #1342, opened 2023-02-02
Check for proper spaces around punctuation


RedX2501:
I see that there's a rule for two spaces after a period but there does not seem to be a rule for the spacing around commas, parenthesis, braces and so on.

Is there no consensus or recommendation around this from your sources?

I have seem many (now I'm not sure if they really are problems) with other writers having spaces around commas, after an opening parenthesis and so on.

arostkowycz15:
I will attempt this issue.

Nytelife26:
> Is there no consensus or recommendation around this from your sources?

This is a purely typographical issue. However, since we already have some checks and sources for typography, this is what I could find on the matter: https://practicaltypography.com/are-two-spaces-better-than-one.html

Logically, whatever we decide upon should be consistent. However, I am inclined to throw this out altogether - there is no best option, and the formatting of one's writing is not traditionally considered prose.

PR #1350, opened 2023-04-17
added check for punctuation spacing. Issue #1342

arostkowycz15:
I added checks for spacing around !?;: as mentioned in issue #1342. Does not include checks for periods or commas as numbers (1,000 or 1.00) add complexity to them.


Nytelife26:
Before I review the rest, why does this change `proselintrc.json` to `proselintrc`?

PR #1368, opened 2024-04-23
ci: pin runner image versions

Nytelife26:
## This relates to...

The key change here is using macos-12 instead of macos-latest. This
prevents the runner from switching to arm64, for which there are no
compiled versions of older Python targets (i.e. 3.8, 3.9).


orgua:
but why also change ubuntu and windows? Going away from "latest" is only a workaround.

Nytelife26:
> but why also change ubuntu and windows? Going away from "latest" is only a workaround.

in general version pinning so breaking changes are not administered automatically is a good way to run things. i will figure out a matrix that tests both, but in the meantime this is fine.

welcome back, by the way

Issue #1334, opened 2022-12-09
Allow inverse `existence_check`s (re: up-goer-five check)


patcon:
Hi there! This tool is neat. Thanks!

I was hoping to create a check for raising issue when any words outside a set were used: (top 1000 english words)
https://splasho.com/upgoer5/phpspellcheck/dictionaries/1000.dicin

Can you suggest the simplest way to go about this? I'm happy to submit a PR when I get around to it :)

Inspiration: https://splasho.com/upgoer5/

vqle23:
Hey, this sounds interesting. I can probably work on this within the next week if you haven't.

Nytelife26:
I can see the potential utility in this, but it feels out of scope for proselint. Can you think of any scenario where this may be useful in checking prose besides in writing tasks that involve special restrictions? I fear this check would be computationally expensive, and unnecessary for the typical use of the program.

patcon:
I feel this sort of processing is most likely to be helpful to people writing linters for things like ESL readers (english-as-second-language) or for scientific communication with very young readers or those with minimal or no high school education.

I don't want to take anyone's time for granted (we should only work on what we enjoy!), but I will admit it feels arbitrary to put these categories of use outside the realm of worth serving. I totally understand if you do not wish to invest your own time, but pls reconsider making a blanket judgement that this is too computationally expensive 🙏🏻🙏🏻🙏🏻🙂

Nytelife26:
Hi, @patcon. I have reconsidered, indeed. Many other linters provide configurations for special environments, and it can't hurt to broaden proselint's utility. Maybe this could be implemented as a plugin, when such a system is available? Alternatively, if you have suggestions for how this could work in a similarly performant manner to what we have now, I would be open to including this in proselint's core.

ferdnyc:
So, with respect to the fine initial work done by @vqle23 in #1351, I have a different implementation of this that I think solves the major issues that came up in that PR's review process.

I started from these basic premises:

1. We have a relatively small universe of words in the "pass" list
2. We'll eventually _need to find the position of every non-matching word_
3. No matter how long the input text is, we're going to end up searching through it all the way to the end.
4. (Tangentially and as a caveat/addendum: At least in the first pass, words shorter than 3 characters aren't worth bothering with.)<sup>1</sup> 

[...]

PR #1369, opened 2024-05-09
checks: Remove duplicate items from lists

ferdnyc:
While reading through one of the `proselint/checks/*/misc.py` files, I happened to notice a duplicate item on its list of matching terms.

So I ran a quick script to look for duplicate items in _all_ of the `misc.py` files, spotted a couple more that way, and removed those as well.

#### Redundancies NOT fixed

Both `cliches/misc.py` and `redundancy/misc.py` contain several more duplicates, by virtue of the same term appearing on multiple **different** lists of cliches/redundancies — those, I didn't touch. This is only about duplicates on the _same_ list.

(Plus, the irony of having redundant terms _in_ the redundancies check is inherently entertaining. Enough to justify leaving those alone anyway.)


Nytelife26 (approved):
good finds, thank you for the contribution.

i like your sense of humour, but i will indeed go through and remove duplicates in other lists as part of this.

ferdnyc:
@Nytelife26 

> i will indeed go through and remove duplicates in other lists as part of this.

Well, my thinking in leaving them was: Where the lists come from different external sources (as is the case in all of those left-over duplicates), how do you choose _which one_ keeps the term, and which one doesn't? (And then the list is no longer in sync with the source.)

ferdnyc:
@Nytelife26 

But if you want the lists of remaining, cross-list duplicates, they are:

`cliches/misc.py`:
```python
        "between a rock and a hard place",
        "conspicuous by its absence",
        "crystal clear",
        "easier said than done",
        "gilding the lily",
        "horns of a dilemma",
        "last but not least",
        "moment of truth",
        "par for the course",
        "thick as thieves",
```

`redundancy/misc.py`:
```python
        ["blend",             ["blend together"]],
        ["Challah",           ["Challah bread"]],
        ["collaborate",       ["collaborate together"]],
        ["combine",           ["combine together"]],
        ["cooperation",       ["mutual cooperation"]],
        ["fundamentals",      ["basic fundamentals"]],
        ["gift",              ["free gift"]],
        ["innovation",        ["new innovation"]],
        ["merge",             ["merge together"]],
        ["mix",               ["mix together"]],
        ["mutual respect",    ["mutual respect for each other"]],
        ["plans",             ["future plans"]],
        ["potable water",     ["potable drinking water"]],
        ["refer",             ["refer back"]],
        ["retreat",           ["retreat back"]],
        ["surrounded",        ["surrounded on all sides"]],
        ["veteran",           ["former veteran"]],
        ["visible",           ["visible to the eye"]],
```

Nytelife26:
> how do you choose _which one_ keeps the term, and which one doesn't?

i am currently waiting to accrue the funds to obtain the sources myself (my university library does not have them unfortunately), which would make things easier, but you're right.

i was considering aggregating the lists so they run by type of check rather than by source, and using comments to keep track, but that might get messy.

i'll come up with something.

ferdnyc:
@Nytelife26 

> i was considering aggregating the lists so they run by type of check rather than by source, and using comments to keep track, but that might get messy.

That's certainly an option. I mentioned in #1334 that in testing my implementation of reverse-existence, I'd factored the lists of terms _out_ of the check functions so that I could access them externally, specifically in the REPL when I was testing/debugging/benchmarking interactively.

That's useful enough that I'd suggest it as an across-the-board change for _all_ of the checking functions — instead of holding their lists internally, move them out to module-level variables which can be accessed from other code.

So, where now the "template" is:

```python
def some_check_fn(args):
    terms = [
      "foo",
      "bar",
      "baz",
    ]
    return some_tool_function(text, terms, blah, blah)
```

My suggestion would be to go with:

```python

SOME_CHECK_TERMS = [
  "foo",
  "bar",
  "baz",
]

def some_check_fn(args):
    terms = SOME_CHECK_TERMS
    return some_tool_function(text, terms, blah, blah)
```

The major advantage there is that you can use `from proselint.checks.checker.misc import SOME_CHECK_TERMS` to get at that list of terms from anywhere in the codebase — from _**any**_ codebase, really, even outside of `proselint`.

But it would also permit combining multiple lists together and de-duplicating them, e.g.:

```python
FIRST_CHECK_TERMS = [...]
SECOND_CHECK_TERMS = [...]

def some_check_fn():
    terms = list(set(FIRST_CHECK_TERMS + SECOND_CHECK_TERMS))

   return some_tool_function(text, terms, blah, blah)
```

Do that, it doesn't matter if the source lists contain redundancies, they'll be dropped in building the final terms list.

Nytelife26:
i like that solution, actually. it could simplify testing, too. i think the goal should eventually be to refactor checks to merely return a list of regex-enabled sequences along with their options, and run the existence check once so it can search the source for everything in as few passes as possible, like what you did with `reverse_existence_check` in #1370.

PR #1351, opened 2023-04-21
Adding Reverse Existence checks Issue #1334

vqle23:
I added a way to ensure that you only words within a certain set as requested in Issue #1334  or as the reporter coin as a "reverse-existence check". Currently, the two sets implemented with this new feature are the top 1000 words set as requested by the initial reporter, but I also added a set of words that elementary kids are taught to learn how to spell. With both sets, I have provided a suite of test cases for quality assurance. However, I did have to modify test_config_flag.py a bit to pass linter tests (some lines were too long, and had to turn a string into a variable on the line above).

PR #1370, opened 2024-05-11
New reverse-existence implementation

ferdnyc:
This PR formally submits the implementation of `proselint.tools.reverse_existence_check` which I describe in https://github.com/amperser/proselint/issues/1334#issuecomment-2100920036. It still builds on the great work @vqle23 did in #1351 (in fact, it builds on the branch from that PR, rebased to current `main` and with my changes added as followup commits.)

A few things have changed from what I described in that issue comment:

1. While the utility function in `proselint.tools` is still named `reverse_existence_check()`, I renamed the actual _checkers_ to `restricted.top1000` and `restricted.elementary`, as I felt that was a less unwieldy title/category name that doesn't sacrifice any accuracy/descriptiveness.
2. I tweaked the word regex slightly, to avoid capturing hyphens or apostrophes on either extreme of the word. (So, my arbitrarily-chosen 3-character minimum length turned out to be prescient, as the regex ended up looking like this: `re.compile(r"\w[\w'-]+\w")`
3. Both versions of the `allowed_word` helper function (case-sensitive and not) are now defined statically in the `tools.py` file. The proper one is selected and bound with a `functools.partial()` at the start of each `reverse_existence_check()` call.

(...And I just noticed, I left some type annotations on the arguments to the helpers, that I'll probably have to take out to appease older Pythons. Bother.) 

Fixes: #1334


proselint/tools.py
```diff
@@ -365,6 +365,52 @@ def existence_check(text, list, err, msg, ignore_case=True, str=False,
...
+def _case_sensitive_allowed_word(permitted: set[str], match: re.Match):
+    """Determine if a match object result is in a set of strings."""
+    matched = match.string[match.start():match.end()]
+    return matched in permitted
+
+
+def _case_insensitive_allowed_word(permitted: set[str], match: re.Match):
+    """Determine if a match object result is in a set, ignoring case."""
+    matched = match.string[match.start():match.end()].lower()
+    return matched in permitted
```
Nytelife26, line 377: these should arguably be merged and use an `ignore_case` flag, like the existence check itself does.

proselint/tools.py
```diff
@@ -365,6 +365,52 @@ def existence_check(text, list, err, msg, ignore_case=True, str=False,
...
+    """Find all words in ``text`` that aren't on the ``list``."""
+    if ignore_case:
+        permitted = set([word.lower() for word in list])
+        allowed_word = functools.partial(
+            _case_insensitive_allowed_word, permitted)
+    else:
+        permitted = set(list)
+        allowed_word = functools.partial(
+            _case_sensitive_allowed_word, permitted
+        )
```
Nytelife26, line 392: if my previous comment is followed, this can also be simplified.
```suggestion
    permitted = set([word.lower() for word in list] if ignore_case else list)
    allowed_word = functools.partial(
        _allowed_word, permitted, ignore_case=ignore_case
    )
```

Nytelife26 (changes requested):
other than this, it looks good. this could be used later as a model for refactoring the way `existence_check` itself works later, actually.

proselint/tools.py
ferdnyc: Hm, not a bad idea, I like it! I'll do that.

proselint/tools.py
ferdnyc: Oh, actually I missed your set-conversion tweak... hm, that feels a bit "fun" to parse (for humans), compared to:
```python
    if ignore_case:
        permitted = set([word.lower() for word in list])
    else:
        permitted = set(list)
```

I'm _open_ to it, just not sure it's necessary to make future code readers work that hard. Thoughts?

ferdnyc:
@Nytelife26 

Oh, one thing I didn't address from the previous PR is [your comment](https://github.com/amperser/proselint/pull/1351#discussion_r1276316348) about the size of the word lists in the code.

I suppose we could move them to a JSON file, and `json.load()` the list in when the check runs. Possibly even a compressed JSON file.

To retain compatibility with zipfile packaging and the like, it'd probably be advisable to use `importlib.resources` to access the data (and make the backport a dependency for older Pythons) rather than relying on there being a filesystem to access. But that's not that big a deal, it's a solved problem at this point.

ferdnyc:
I just sketched out a quick attempt at offloading the top1000 word list into a JSON file. Uncompressed, `top1000.json` would be 8.9k, and reduce the size of `top1000.py` from 14k to 850 bytes. (That's including the code to load the word list from the file.)

ferdnyc:
`elementary.json` would be 7.5k, and reduce `elementary.py` from 12k to around the same 850 bytes.

proselint/tools.py
Nytelife26: i have never really considered the readability of inline if statements for Python programmers, i just know ternaries are a popular construct in other languages i use, like Rust and TypeScript. additionally, ruff has [a rule](https://docs.astral.sh/ruff/rules/if-else-block-instead-of-if-exp/) (derived from flake8-simplify) that prefers ternaries, so i would imagine they're not uncommon in Python as well.

Nytelife26:
> it'd probably be advisable to use `importlib.resources` to access the data (and make the backport a dependency for older Pythons)

since `importlib.resources` was added in Python 3.7, and the oldest supported version of Python is 3.8, we don't need to worry about backports. i would be comfortable with using `importlib` if you think it'd be more compatible for packaging.

proselint/tools.py
ferdnyc: I tend to reserve them for situations where there's no other choice, like in list comprehensions. (Which can be downright squirrely to read under any circumstances; see my last commit that adds some indentation to try and make my whopper at least _semi_-readable.) I guess it all comes down to personal comfort; I know people who dislike code like this:
```python
if something:
    return one thing
return the other
```
...and always want the explicit `else:` in there. And then OTOH there are linters that will _flag_ the explicit `else:` as unnecessary (which of course it _is_, technically).

`¯\_(ツ)_/¯`

ferdnyc:
@Nytelife26 Unfortunately, the modern interface to `importlib.resources` (`importlib.resources.file()`) didn't come along until Python 3.9. But there's [the `importlib-resources` backport package](https://pypi.org/project/importlib-resources/) that can pull it in for Python 3.8.

It's just a matter of adding a dependency (I can never remember the syntax..):
```text
importlib-resources = "^6.0"; python_version < "3.9"
```

ferdnyc:
Hmm, turning the list into a `.csv` file with zero structure whatsoever makes `top1000.csv` 5.9k, as opposed to the JSON file's 8.9k.

That's for a file that looks like _this_, which I hate (but I don't have to read it, Python does):
```csv
a,able,about,above,accept,across,act,actually,add,admit,afraid,after,afternoon,again,against,age,ago,agree,ah,ahead,air,all,allow,almost,alone,along,already,alright,also,although,always,am,amaze,an,an
```

The JSON file looks like this, technically I could squeeze it tighter by telling it to lose the spaces between each item:

```json
["a", "able", "about", "above", "accept", "across", "act", "actually", "add", "admit", "afraid", "after", "afternoon", "again", "against", "age", "ago", "agree", "ah", "ahead", "air", "all", "allow",
```

Going with CSV also means dealing with `csv.py` and the `csv.reader()`, which as standard library modules go is abominable. `json.load()` is simplicity itself. But, it's just 2-4 lines of code either way, so NBD. And the file _is_ a lot smaller.

ferdnyc:
> It's just a matter of adding a dependency (I can never remember the syntax..):
> 
> ```
> importlib-resources = "^6.0"; python_version < "3.9"
> # Wait, actually I think it's...
> importlib-resources = "^6.0; python_version < 3.9"
> ```

I was doubly wrong, with poetry it's apparently:

```ini
importlib-resources = { version = "^6.0", python = "<3.9" }
```

Nytelife26:
> That's for a file that looks like _this_, which I hate (but I don't have to read it, Python does)

using newlines as field delimiters instead of commas would improve readability and make diffs easier, without changing file size, if that helps.

Nytelife26:
it seems that, at least on my system, `csv.reader` expects all entries to be quoted by default. i'll have to look into that.

Nytelife26:
sort of. it still produces a list of lists (on my system at least), which should be trivial to flatten, but it's worth noting.

ferdnyc:
> sort of. it still produces a list of lists (on my system at least), which should be trivial to flatten, but it's worth noting.

Yeah, I ended up having to do _this_ nonsense which is... whatever. I hate `csv.py` with a passion.

```python
with files(proselint).joinpath(_CSV_PATH).open('r') as data:
    reader = csv.reader(data)
    wordlist = list()
    for row in reader:
        wordlist.extend(row)
```

Nytelife26:
not many releases have been conducted since Professor Suchow dropped all communication, so i am not entirely sure either. i presume it works, since the configuration file always has, but i will test it locally soon.

Nytelife26:
it would be nice if python had a flatten function, like pretty much every other language i use, that's for sure.

my latest commits should ease your pain somewhat.

proselint/checks/restricted/top1000.py
```diff
@@ -0,0 +1,35 @@
...
+
+with files(proselint).joinpath(_CSV_PATH).open('r') as data:
+    TOP1000_WORDS = data.read().split()
+
+
+@memoize
+def check(text):
+    """Check the text."""
+    err = "restricted.top1000"
+    msg = "'{}' is not in the top 1,000 most popular words."
```
Nytelife26, line 33:
```suggestion
    msg = "'{}' is not in the top 1000 most common words."
```

Nytelife26:
i'm ready to merge this if you have no further comments, suggestions or changes to make.

ferdnyc:
@Nytelife26 All looks good to me, thanks for all your help with this!

PR #1371, opened 2024-05-13
refactor: modernisation of proselint

Nytelife26:
this is a follow on from #1361. credit to @orgua for most of the initial work here.

TODO:
- [ ] Make scientific.numbers and scientific.misc pass tests
- [ ] Resolve merge conflicts with base
- [ ] Resolve TODOs
- [ ] Work on discussion from #1369

Issue #1362, opened 2024-01-22
[feature-proposal] never mind


orgua:

Nytelife26:
I like where this is going so far. Here are my additional proposals for the check structure

Out of scope considerations: cursing, security, links, spelling. Each has merits, but they feel out of place in the context of the goal of proselint.

Condensing considerations: 
- `lgbtq`, `sexism` and similar can be combined. This category might be named something like `social_awareness`
- `airlinese`, `corporate_speak` and similar can be combined. This category might be named something like `industrial_language`. Optionally, `jargon` may be added to this.
- `mondegreens` and `malapropisms` are commonly grouped together in analysis.
- Punctuation spacing and `typography` should be together, too. Spacing is a purely typographic matter.

I also liked your previous suggestion of flattening categories with only one file. It might be worth making a standard for check naming, while we're at it.

Let me know what you think.

PR #1361, opened 2024-01-05
modernization of proselint

orgua:
Hello,

I put some days into improving the codebase. A lot of open issues should be solved by this. BUT changes are far more substantial as planned - so I wanted to ask how to proceed.

This repo is not very active in the last years. I would offer to maintain the project for the time being and also be open for guiding input. Some of the open Pull Request would be beneficial to the project. I also got plans for new features, improving ease-of-use, more tests for correctness and other optimizations.

I'm open for a chat. Mail is in `setup.cfg` of one of my [projects](https://github.com/orgua/shepherd-datalib).

If there are other plans for the project I would probably fork and start fresh - this is not my preferred solution.

## Biggest Changes

- featureset of py38
- cleaner and faster code
- fixed bugs, undefined behavior, broken legacy code
- some checks did not work as intended
- multiprocessed parallelization and other optimizations

## Benchmark-Comparison

- custom set of files, same hardware

### Windows

Proselint-main (uncached & cached)

- Found 104 lint-warnings in 47.579 s
- Found 104 lint-warnings in 0.878 s

Proselint-modernized

- Found 108 lint-warnings in 39.930 s (12 files, 617.45 kiByte) -> serialized
- Found 108 lint-warnings in 13.164 s (12 files, 617.45 kiByte) -> serial files, parallel checks
- Found 108 lint-warnings in 9.931 s (12 files, 617.45 kiByte)  -> global check-executor
- Found 108 lint-warnings in 0.011 s (12 files, 617.45 kiByte)  -> cached

### Linux / WSL

Proselint-main (uncached & cached)

- Found 104 lint-warnings in 37.041 s
- Found 104 lint-warnings in 0.771 s

Proselint-modernized

- Found 108 lint-warnings in 34.521 s (12 files, 617.45 kiByte)
- Found 108 lint-warnings in 8.248 s (12 files, 617.45 kiByte)
- Found 108 lint-warnings in 6.098 s (12 files, 617.45 kiByte)
- Found 108 lint-warnings in 0.044 s (12 files, 617.45 kiByte)

## Breaking changes

- cli arg `--time` -> `--benchmark`, `-b`
- cli arg `--debug` ->  `--verbose`, `-v`
- exit code now returns number of errors
- cli arg `--output-format` controls old flags json, compact
- py >= 3.8
- api-changes ->  web_scripts & plugins untested
- config adjustments (1 test removed, 1 added)
	- "misc.metaconcepts": False,  # TODO: remove, was duplicate of scare_quotes

[...]

PR #1372, opened 2024-05-16
Update er_or.py - remove redundant 'promoter'.

davidhicks980:
Please reject this PR if this is not an issue.

### DIFF
--- a/proselint/checks/spelling/er_or.py
+++ b/proselint/checks/spelling/er_or.py
@@ -31,7 +31,6 @@ def check(text):
         ["mortgagor",           ["mortgager"]],
         ["persecutor",          ["persecuter"]],
         ["promoter",            ["promotor"]],
-        ["promoter",            ["promotor"]],
         ["purveyor",            ["purveyer"]],
         ["requester",           ["requestor"]],
         ["reviser",             ["revisor"]],


Nytelife26:
@ferdnyc it seems evident we may not have caught them all. i'll have to review more comprehensively

ferdnyc:
@Nytelife26 Mmm, my scan was admittedly less-than-comprehensive, and would've easily been thrown off by (for example) differences in indentation or formatting. (It used `sort` and `uniq` in the shell to detect dupes.)

ferdnyc:
@orgua 

> keep in mind that there is a massive overhaul coming 

That's a good point; looking at the changes in #1361, I don't see anything that's _incompatible_ with extracting the lists as importable entities, but it also doesn't **do** that, and I'm still a strong proponent of taking that step. The benefits of being able to `from proselint.checks.foo.bar import BAR_TERMS` and manipulate the list of terms as a discrete object can't be overstated. Easy deduplication, easy statistical analysis, etc, etc.

Making that change outside of 1361 would certainly be incredibly disruptive, and make it far harder to merge that PR. Doing it as a post-1361 change would largely require waiting for it to be merged, first. (Assuming one wants to preserve one's sanity by **not** coding against a moving target like an open PR.)

By the same token, even merging _this_ PR slightly disrupts #1361; if the intent is to get that in, it might be easier to suspend **all** other changes and focus on getting just the big'un over the finish line. IMHAndUnsolicitedO.

orgua:
yes, disruptive. I agree. Unfortunately @Nytelife26 is still merging smaller PRs before that large overhaul and thus making rebasing an operation that is easily susceptible to human error and hard to verify.

Arguments against the "global" list are: 
- it was slightly slower in my tests
- some list are generated dynamically, so this wouldn't be a universal approach (but that's fixable by using the function-output)

The testsuite is already improved and checks for some duplication, but not entries between different checks. Some doubles might be wanted though. When you delete them in one test and disable the other via configuration you would miss warning. The other approach would be to do it JIT for performance reasons. But also that might be more expensive than just rechecking these doubled entries.

Note: You can access list-variables that are inside functions in imported modules if you just want to analyze things.

But beside that it would be beneficial to further overhaul the check-files - add more machine readable metadata like: enabled, preview, language, ... This was proposed by me in #1362 and would get rid of the hardcoded config-file (code duplication). Additional to that change it would be possible to also add a checktype-metadata and just have that list that you mentioned. At least it would be possible for the most common existence check.

Nytelife26:
> @ferdnyc keep in mind that there is a massive overhaul coming (#1361 and #1371). File structure and check-design are changing significantly. 

if you plan on sticking around to continue working on it with me, i can close #1371 and we can pick up where we left off. it's up to you whether or not you wish to continue on this, or whether you'd prefer to leave the rest to me. in any case, i greatly appreciate the efforts you've put into it. you've been massively helpful in setting out a clear, definitive way for proselint to have a future.

ferdnyc:
Oh, man, I had a very long reply typed up here that naturally got lost in a Chrome crash. But maybe that's for the best, you all didn't really need to be subjected to all of that and this conversation should really be in an Issue or Discussion anyway. But I will just say:

@orgua 
> Arguments against the "global" list are:
> 
> * it was slightly slower in my tests
> * some list are generated dynamically, so this wouldn't be a universal approach (but that's fixable by using the function-output)

On the speed issue, I can't believe there would be any significant difference, and my own tests don't bear that out. It's a reference to a list, which is immediately being passed to another function. Whether that reference points to a function-scope symbol or a module-scope symbol can't possibly make any appreciable difference.

The only scenario I can come up with, where module-level variables might be appreciably slower, is when you're running with a lot of checks disabled... and even then, only _if_ proselint is still importing **all** of the check modules (even the disabled ones). But in that case I'd say that the right fix is to make the import code smarter and avoid importing anything it doesn't need.

Generated lists are possibly the _most_ important to hoist out of the function, I'd say, because why should they be regenerated every time it's called? The resulting lists being generated are static in nearly all cases, so it makes sense to store them somewhere for reuse.

In fact, for some of the generated lists, where the actual generation is just a matter of convenience/laziness, if speed is _really_ a concern then those checks could be optimized by simply running the generator function in a terminal and dumping the **output** into the code as a static list.

(I'm talking about things like in `security/password.py`...):
https://github.com/amperser/proselint/blob/94a1eea991ffc171a28115523876add890f78225/proselint/checks/security/password.py#L23-L31

...I don't think the runtime of those kinds of operations is significant enough to be worth worrying about (not even the _significantly_ bigger list generated in `misc/waxing.py`), but the option is there.

> Note: You can access list-variables that are inside functions in imported modules if you just want to analyze things.

How? If there's a way other than by digging around in its `__code__` object (or having the `inspect` module do it for you), I don't know it. And `__code__` spelunking is tricky enough to do by hand, even harder to do programmatically. The symbols end up in weird and not-always-predictable places.

Plus, that **still** doesn't get you access to the _generated_ lists, because the function has to be executed for those variables to exist.

orgua:
some minimal feedback: 

- accessing global vs local variables has different performance, even in newer python versions - but I agree, it does not matter in this case (for your mentioned static lists)
- generated global lists get processed when the module is loaded, these local ones process when the function is executed. So the discovery process of proselint is cheaper with local generated variables. I would still prefer to keep the discovery process and not hardcode the list of available test.
- These generated lists should be best assembled but during compile time, like constexpr. Python hasn't something even close to it. Porting the core functionality to rust was my far stretched goal ...
- my proposed datastructure behaves like a class, has accessible variable and might be lazy evaluated during discovery - good for the generated content. otherwise the other major speedups should justify the overhead.

Nytelife26:
@orgua 

> * These generated lists should be best assembled but during compile time, like constexpr. Python hasn't something even close to it. Porting the core functionality to rust was my far stretched goal ...

i was considering this for a while. even typescript would be an improvement if you ask me. the major issue there is it's a large project and a rust rewrite would be a proportional undertaking (and import dynamics work a lot differently).

while you're here, would you mind answering my previous comment?

> if you plan on sticking around to continue working on it with me, i can close #1371 and we can pick up where we left off. it's up to you whether or not you wish to continue on this, or whether you'd prefer to leave the rest to me.

it would be helpful to know if it's worth proceeding with my own efforts or if i should wait for you to have more time so we can continue collaborating properly.

orgua:
proselint is large? i wouldn't even call it medium :) There are some base algos and the rest is gluecode and definitions of checks.

@Nytelife26 do you have access to your last mail address? there should be two mails waiting. i might have time - but lets discuss that elsewhere.

Nytelife26:
> @Nytelife26 do you have access to your last mail address? there should be two mails waiting.

my apologies, i tend not to check my university email much outside of term dates. i have replied, so hopefully progress awaits.

Nytelife26:
@orgua just a ping to make sure you're aware we've set a time. let me know if you can make it.

PR #1398, opened 2025-08-18
refactor: implement iterators for check results

Nytelife26:
- Use iterators for check types and flags
- Enable early halting for flag conditions

This is primarily focused on ergonomics for future changes, but it also provides a small performance increase (~10ms for a full demo run) over main.

### DIFF
--- a/proselint/tools.py
+++ b/proselint/tools.py
@@ -32,15 +32,11 @@ def line_and_column(text, position):
 
 
 def lint(
-    content_or_file: Union[str, FileIO],
+    source: Union[str, FileIO],
     config: Config = DEFAULT,
 ) -> list[LintResult]:
     """Run the linter on the input."""
-    content = (
-        content_or_file
-        if isinstance(content_or_file, str)
-        else str(content_or_file.read())
-    )
+    content = source if isinstance(source, str) else str(source.read())
 
     return sorted(
         islice(


proselint/registry/checks/__init__.py
```diff
@@ -90,61 +94,57 @@ class CheckFlags(NamedTuple):
...
-                message=last_message,
-                replacements=None,
+        return chain(
+            islice(results, limit - 1),
+            (
+                CheckResult(
+                    start_pos=result.start_pos,
+                    end_pos=result.end_pos,
+                    check_path=result.check_path,
+                    message=f"{result.message} Also found elsewhere.",
```
drainpixie, line 115: Minor nitpick, unless it makes sense in this specific context.
```suggestion
                    message=f"{result.message} also found elsewhere.",
```

proselint/registry/checks/__init__.py
```diff
@@ -90,61 +94,57 @@ class CheckFlags(NamedTuple):
...
+        return (
             CheckResult(
-                start_pos=results[0].start_pos,
-                end_pos=results[0].end_pos,
-                check_path=results[0].check_path,
-                message=results[0].message + f" Reached {ppm:.0f} ppm.",
+                start_pos=result.start_pos,
+                end_pos=result.end_pos,
+                check_path=result.check_path,
+                message=f"{result.message} Surpassed {threshold} ppm.",
```
drainpixie, line 136: Ditto as above

proselint/tools.py
```diff
@@ -31,43 +32,36 @@ def line_and_column(text, position):
 
 
 def lint(
-    input_file: Union[str,
-    FileIO],
+    input_file: Union[str, FileIO],
```
drainpixie, line 35:
```suggestion
    input: Union[str, FileIO],
```
Just think this would be neater, `input` isn't guaranteed to be a file and it's more generic
This is assuming there is programmatic capability and it's not just redundant

proselint/tools.py
```diff
@@ -31,43 +32,36 @@ def line_and_column(text, position):
...
     config: Config = DEFAULT,
-    *,
-    debug: bool = False,
 ) -> list[LintResult]:
     """Run the linter on the input file."""
```
drainpixie, line 38:
```suggestion
    """Run the linter on the input."""
```
Ditto

drainpixie (changes requested):

proselint/registry/checks/__init__.py
Nytelife26: The idea here is that check messages are supposed to end with a period anyway. Take, for instance, `archaism`: "'{}' is archaic." would become "'{}' is archaic. Also found elsewhere."

drainpixie: Noted, I feel like it'd be better to do this in this function itself rather than rely on the check's messages, though that'd be a lot of work. If necessary, we could always add validation within the check constructor itself.

proselint/tools.py
drainpixie: `source` could also work

drainpixie (approved):

PR #1404, opened 2025-09-01
fix(lint): add flag to prevent filtering quotes in lintfile

drainpixie:
- added an `allow_quotes` flag to `CheckFlag`s to allow proper matching of quotes
- removed dead code (`find_quoted_spans`)

### DIFF
--- a/proselint/tools.py
+++ b/proselint/tools.py
@@ -170,8 +170,8 @@ class LintFile:
                     for check in registry.get_all_enabled(config["checks"])
                     for result in check.check_with_flags(self.content)
                     if (
-                        not self.is_quoted_pos(result.start_pos)
-                        or check.flags.allow_quotes
+                        check.flags.allow_quotes
+                        or not self.is_quoted_pos(result.start_pos)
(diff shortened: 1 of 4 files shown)


proselint/tools.py
```diff
@@ -174,7 +169,11 @@ def lint(self, config: Config = DEFAULT) -> list[LintResult]:
                     )
                     for check in registry.get_all_enabled(config["checks"])
                     for result in check.check_with_flags(self.content)
-                    if not self.is_quoted_pos(result.start_pos)
+                    if (
+                        not self.is_quoted_pos(result.start_pos)
+                        or check.flags.allow_quotes
```
Nytelife26, line 174: The static condition should come first here.
```suggestion
                        check.flags.allow_quotes
                        or not self.is_quoted_pos(result.start_pos)
```

Nytelife26 (changes requested):
Other than a tiny nitpick, looks good.

Nytelife26 (approved):

PR #1407, opened 2025-09-01
perf(tools): use bisect for boundary checks

Nytelife26:


proselint/tools.py
```diff
@@ -177,16 +179,13 @@ def _line_bounds(self) -> tuple[int, ...]:
...
-        )
-        return (len(self.line_bounds) - line_delta, position - bound)
+        bound_idx = bisect_left(self.line_bounds, position)
+        return (bound_idx, position - self.line_bounds[bound_idx - 1])
 
     def is_quoted_pos(self, position: int) -> bool:
         """Determine if the content position falls within quotes."""
-        return any(start <= position < end for start, end in self.quote_spans)
+        # NOTE: since bounds are paired, odd insertions are always within a span
+        return bisect_left(self.quote_bounds, position) % 2 == 1
```
Nytelife26, line 188: This technically makes the new `allow_quotes` flag redundant because `bisect_left` is equivalent to a less-than operation, but I would be in favour of keeping it to make things explicit and prevent the O(log *n*) computation.

Nytelife26:
@drainpixie This is ready to go :)

drainpixie (approved):

PR #1413, opened 2025-09-02
fix: rectify capitalisation in preferred forms

drainpixie:


Nytelife26 (changes requested):
Looks great other than these. I'm especially impressed with the work you've done in the test suite, providing infrastructure for future case insensitive testing.

tests/test_check_types.py
```diff
@@ -164,6 +186,69 @@ def test_preferred_s_smoke(
...
+    )
+
+    text = " ".join(items.keys())
+    check_type = types.PreferredFormsSimple(items=items, padding=padding)
+    results = list(check_type.check(text, check))
+
+    assert all(result.replacements in items.values() for result in results)
+    for i, result in enumerate(results):
+        key = list(items.keys())[i]
+        expected_replacement = items[key]
```
Nytelife26, line 208: Since you don't require the custom bisect and accumulation logic here, it'd be neater to do something like this.

```suggestion
    for result, (original, replacement) in zip(results, items.items()):
```

tests/test_check_types.py
```diff
@@ -164,6 +186,69 @@ def test_preferred_s_smoke(
...
+    )
+
+    check_type = types.PreferredFormsSimple(items=normalised, padding=padding)
+    results = list(check_type.check(text, check))
+    assert all(result.replacements in normalised.values() for result in results)
+
+    text_keys = [
+        k.swapcase() if i % 2 == 0 else k.upper()
+        for i, k in enumerate(items.keys())
+    ]
```
Nytelife26, line 240: This is duplicate code. It should be constructed first, and then the test data should be created by calling `" ".join` over it.

tests/test_check_types.py
```diff
@@ -164,6 +186,69 @@ def test_preferred_s_smoke(
...
+        ignore_case=True,
+    )
+
+    text = "  ".join(
+        k.swapcase() if i % 2 == 0 else k.upper()
+        for i, k in enumerate(items.keys())
+    )
+
+    check_type = types.PreferredFormsSimple(items=normalised, padding=padding)
+    results = list(check_type.check(text, check))
```
Nytelife26, line 234: If the only results evaluated are with the `normalised` dictionary, construct the `check` with this in the first place, and use `check.check`.

tests/test_check_types.py
```diff
@@ -164,6 +186,69 @@ def test_preferred_s_smoke(
...
+    assert all(result.replacements in normalised.values() for result in results)
```
Nytelife26, line 235:
```suggestion
```
This is guaranteed by the assertion loop.

tests/test_check_types.py
```diff
@@ -164,6 +186,69 @@ def test_preferred_s_smoke(
...
+
+    text_keys = [
+        k.swapcase() if i % 2 == 0 else k.upper()
+        for i, k in enumerate(items.keys())
+    ]
+
+    for i, result in enumerate(results):
+        original_key = list(items.keys())[i]
+        text_key = text_keys[i]
+        expected_replacement = normalised[original_key.lower()]
```
Nytelife26, line 245: You can do similar tidying here as with the previous function.

Nytelife26 (changes requested):

Nytelife26 (approved):

PR #1414, opened 2025-09-02
fix: check venery-waxed return types

drainpixie:


proselint/checks/terms/venery.py
```diff
@@ -72,14 +73,14 @@
...
 
 
-def _check_venery(text: str, check: Check) -> list[CheckResult]:
+def _check_venery(text: str, check: Check) -> Iterator[CheckResult]:
     """Check the text."""
     if not search(
         Padding.SAFE_JOIN.format("|".join(GENERIC_TERMS)), text, check.re_flag
     ):
-        return []
+        return
```
Nytelife26, line 81: This should `return iter(())`.

proselint/checks/terms/venery.py
```diff
@@ -72,14 +73,14 @@
...
 
-    return types.PreferredFormsSimple(
+    yield from types.PreferredFormsSimple(
```
Nytelife26, line 83: This should be changed back to `return`.

Nytelife26 (changes requested):
Same feedback for waxed.

Nytelife26 (approved):

PR #1416, opened 2025-09-02
ci: use nix for workflows

### LINKED ISSUE #1382: [ci] use nix for workflows
Posted by @drainpixie in #1381 ([comment](https://github.com/amperser/proselint/pull/1381#issuecomment-2949567156))
> Once the Nix part of this project is more fleshed out, using Nix for the CI rather than messy YAML files should be considered.

Reproducible workflows will be valuable as the project progresses. This is of negligible importance for now, but it would be a good goal for the near future.

This should incorporate #1201.

drainpixie:
Closes #1382.


.github/workflows/ci-lint-test.yml
```diff
@@ -1,50 +1,83 @@
 name: "CI: Lint & Test"
-on: [push, pull_request]
+
+on: [push, pull]
```
Nytelife26, line 3: According to the [Workflow syntax page](https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#on), this should still be `pull_request`.

Nytelife26 (changes requested):
Looks good so far, other than this one detail.

.github/workflows/ci-lint-test.yml
drainpixie: My bad, good catch

Nytelife26 (commented):
This looks great. Thank you for all the hard work! Hopefully the tokens work correctly when the changes land on main, but if not I'll be able to troubleshoot that. After the `pull_request` trigger and `# TODO: Notify` are removed, I can approve this if you have nothing left to do.

PR #1421, opened 2025-09-03
fix(registry)!: remove results_limit

### LINKED ISSUE #1324: Print out every occurrence of an issue
I'd love to use proselint (especially with the VSCode extension) but one thing blocking me is that it will only ever print out the first occurrence of an error.

For example, if you give this input:

```
This is a very cool and nice test; it is very awesome.

I'm writing some very cool text and it will be very cool.
```

This is the output:

```
test.md:1:11: weasel_words.very Substitute 'damn' every time you're inclined to write 'very'; your editor will delete it and the writing will be just as it should be. Found 4 times elsewhere.
```

This is not very helpful if, for example, you are writing a 50,000-word novel and sometimes characters speak and say things like "very," but you still want to catch usages of "very" in other scenarios.

This makes the VSCode extension basically useless since it will only ever highlight the first occurrence.

I would love an option like `--list-all` that would not collapse multiple occurrences like this.

Nytelife26:
## Relevant issues

Resolves #1324.

### :warning: Breaking

- registry: remove `results_limit` from `CheckFlags`

### DIFF
--- a/proselint/checks/cliches/hell.py
+++ b/proselint/checks/cliches/hell.py
@@ -14,13 +14,12 @@ Never use the phrase 'all hell broke loose'.
 
 """
 
-from proselint.registry.checks import Check, CheckFlags, types
+from proselint.registry.checks import Check, types
 
 check = Check(
     check_type=types.ExistenceSimple(pattern="all hell broke loose"),
     path="cliches.hell",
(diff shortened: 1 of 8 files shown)


Nytelife26:
@drainpixie This is ready for review.

Issue #703, opened 2017-04-26
misc.but Gives incorrect line numbers (other tests give correct ones)


tringham:
The misc.but test gives line numbers inconsistent with other tests

Test file:
Very unique ison line one

Hey!!! Is on lne three

But, wait, line five is wrong

Well, so it is - but line 7 is fine

json:

{
	"data": {
		"errors": [{
			"check": "uncomparables.misc",
			"column": 1,
			"end": 13,
			"extent": 12,
			"line": 1,
			"message": "Comparison of an uncomparable: 'very unique ' is not comparable.",
			"replacements": null,
			"severity": "warning",
			"start": 1
		}, {
			"check": "weasel_words.very",
			"column": 2,
			"end": 6,
			"extent": 4,
			"line": 1,
			"message": "Substitute 'damn' every time you're inclined to write 'very;' your editor will delete it and the writing will be just as it should be.",
			"replacements": null,
			"severity": "warning",
			"start": 2
		}, {
			"check": "hyperbolic.misc",
			"column": 1,
			"end": 35,
			"extent": 7,
			"line": 3,
			"message": "'Hey!!!' is hyperbolic.",
			"replacements": null,
			"severity": "warning",
			"start": 28
		}, {
			"check": "leonard.exclamation.multiple",
			"column": 5,
			"end": 34,
			"extent": 2,
			"line": 3,
			"message": "Stop yelling. Keep your exclamation points under control.",
			"replacements": null,
			"severity": "warning",
			"start": 32
		}, {
			"check": "misc.but",
			"column": 1,
			"end": 56,
			"extent": 5,
			"line": 4,
			"message": "No paragraph should start with a 'But'.",
			"replacements": null,
			"severity": "warning",
			"start": 51
		}]
	},
	"status": "success"
}

PR #1423, opened 2025-09-03
fix: simplify but check and properly count lines

drainpixie:
- Split file in two to get individual offsets
- Varying-width match beginnings make offsets inconsistent

Refs: #703, #1736

### DIFF
--- a/proselint/checks/misc/but.py
+++ b/proselint/checks/misc/but.py
@@ -17,10 +17,11 @@ Paragraphs should not start with certain bad words.
 from proselint.registry.checks import Check, types
 
 check = Check(
-    check_type=types.Existence(items=(r"(?:^|([\.\!\?]\s*))But",)),
+    check_type=types.ExistenceSimple(r"(?:\n\n|^\n)But"),
     path="misc.but",
     message="No paragraph should start with a 'But'.",
     ignore_case=False,
+    offset=(2, 0)
 )
 
 __register__ = (check,)


proselint/checks/misc/but.py
```diff
@@ -14,13 +14,31 @@
...
+        )
+
 
 check = Check(
-    check_type=types.Existence(items=(r"(?:^|([\.\!\?]\s*))But",)),
+    check_type=_check_but_paragraphs,
     path="misc.but",
     message="No paragraph should start with a 'But'.",
     ignore_case=False,
+    offset=(0, 0)
```
Nytelife26, line 41:
```suggestion
```
This is the default.

proselint/checks/misc/but.py
```diff
@@ -14,13 +14,31 @@
...
 
-from proselint.registry.checks import Check, types
+import re
+from collections.abc import Iterator
+
+from proselint.registry.checks import Check, CheckResult
+
+
+def _check_but_paragraphs(text: str, check: Check) -> Iterator[CheckResult]:
+    pattern = r"(^|^\n|\n\n)But\b"
```
Nytelife26, line 24:
```suggestion
    pattern = r"(^\n|\n\n)But\b"
```
The content will never start without a linebreak, due to the content wrapping in `LintFile`.

Also, consider moving this outside of the function into a "constant" wrapped with `re.compile`.

proselint/checks/misc/but.py
```diff
@@ -14,13 +14,31 @@
...
+    for m in re.finditer(pattern, text):
+        count = m.group(0).count("\n", 0, 2)
+        yield CheckResult(
+            start_pos=m.start() + count + check.offset[0],
+            end_pos=check.offset[1],
+            check_path=check.path,
+            message=check.message,
+            replacements=None,
+        )
```
Nytelife26, line 33: This is equivalent to running an `ExistenceSimple` check. It would be preferable to return an iterator expression regardless, due to performance improvements for Python 3.13+.

proselint/checks/misc/but.py
```diff
@@ -14,13 +14,31 @@
...
+from proselint.registry.checks import Check, CheckResult
+
+
+def _check_but_paragraphs(text: str, check: Check) -> Iterator[CheckResult]:
+    pattern = r"(^|^\n|\n\n)But\b"
+    for m in re.finditer(pattern, text):
+        count = m.group(0).count("\n", 0, 2)
+        yield CheckResult(
+            start_pos=m.start() + count + check.offset[0],
+            end_pos=check.offset[1],
```
Nytelife26, line 29: This should use `match.end()`, or `result.end_pos` if you convert to `ExistenceSimple`.

Nytelife26 (changes requested):

proselint/checks/misc/but.py
drainpixie: It will, in tests, I was considering whether to modify the tests to just "mock" a lint file by prepending a \n or not, but I'll let you decide

tests/examples.py
```diff
@@ -190,8 +190,8 @@
         (
             'But I never start with the word "but".',
             "But why are you like that.",
-            "This is cool! But that not so much.",
-            "Is this cool? But that not so much.",
+            "This is a paragraph.\n\nBut this starts a new paragraph",
+            'Some text here.\n\nBut wait, there\'s more',
```
Nytelife26, line 194:
```suggestion
            "Some text here.\n\nBut wait, there's more",
```

Issue #306, opened 2016-03-07
checks are improperly triggered by hyphenated words.


joeyo:
It seems that proselint is using hyphens as word boundaries. This can lead to false positives. For example, this phrase: 

>  ... a 4th order low-pass with a roll-off of 24 db/octave ... 

triggered _ganer.preferred_forms_, because "off" is preferred to "off of". Roll-off, of course, is a completely unrelated compound word. While this particular example is fairly jargony, there's potential for other false positives of this class.

suchow:
Thanks for reporting this, much appreciated — having a low false alarm rate is a priority.

PR #1428, opened 2025-09-10
feat: add stricter padding for off-of check

drainpixie:
## Relevant issues

- #306 

## Brief

Prevent false positives due to words glued together by hyphens and apostrophes

## Changes

- Added test cases for #306 
- Added a `STRICT_WORDS_IN_TEXT` that does not include hyphens or apostrophes
- Fixed the `misc.preferred_forms` check


Nytelife26:
Unfortunately, future plans include converting as many checks as possible to a Thompson NFA, so alternative solutions would be preferable. We may have to look into implementing character-consuming paddings to resolve that (e.g. `\s{}\s`), but this is a good start for now.

PR #1436, opened 2025-09-16
perf(checks)!: use dual engine system

Nytelife26:
## Brief

Here, we achieve a performance increase of ~33% against a single file (benchmarked on my system) by using a hybrid match engine approach, similar to the technique planned on #1371. This also reduces the time taken to lint further files by ~75% on my system.

Further improvements could be made by swapping out Python's re for PCRE2, but the performance difference is negligible in contrast with the dependency growth due to the small number of checks requiring `Engine.FANCY`.

## Changes

- **BREAKING:** Remove `ignore_case` from `Check`, replacing it with a `matcher` containing `RegexOptions`
- Implement a dual-engine wrapper over RE2 and Python's re under `proselint.registry.checks.engine`
- Cache compiled patterns
- Add an engine selector to `Check.matcher`, so `Engine.FANCY` (re) can be used for backreferences and lookarounds
- Update tests


proselint/registry/checks/__init__.py
```diff
@@ -160,3 +123,6 @@ def check(self, text: str) -> Iterator[CheckResult]:
     def check_with_flags(self, text: str) -> Iterator[CheckResult]:
         """Apply the check over `text`, including specified `CheckFlags`."""
         return self.flags.apply(self.check(text), len(text))
+
+
+__all__ = ("Check", "CheckFlags", "CheckResult", "LintResult", "Padding")
```
drainpixie:
```suggestion
__all__ = ("Check", "CheckFlags", "CheckResult", "LintResult")
```

proselint/registry/checks/engine.py
```diff
@@ -0,0 +1,322 @@
...
+
+    case_insensitive: bool
+    multi_line: bool
+
+    def __init__(
+        self, *, case_insensitive: bool = True, multi_line: bool = False
+    ) -> None:
+        """Set the options."""
+        self.case_insensitive = case_insensitive
+        self.multi_line = multi_line
```
drainpixie: Can't this just be a `@dataclass`? and being cached, it can probably be frozen & hashable, too

proselint/registry/checks/engine.py
```diff
@@ -0,0 +1,322 @@
...
+    """A match set based on the `Fast` engine."""
+
+    engine: Engine
+    padding: Padding
+    _anchor: Anchor
+    _patterns: tuple[str, ...]
+    _set: re2.Set
+
+    @cache
+    @staticmethod
```
drainpixie: Ditto

proselint/registry/checks/engine.py
```diff
@@ -0,0 +1,322 @@
...
+    def _construct_set(
+        engine: Engine,
```
drainpixie:
```suggestion
        engine: Fast,
```

drainpixie (changes requested):

drainpixie (dismissed):

drainpixie (dismissed):

PR #1440, opened 2025-09-29
feat: disable checks

drainpixie:

### DIFF
--- a/proselint/config/__init__.py
+++ b/proselint/config/__init__.py
@@ -1,9 +1,10 @@
 """Configuration for proselint."""
 
 import json
+from collections.abc import Hashable
 from importlib.resources import files
 from pathlib import Path
-from typing import TypedDict
+from typing import TypedDict, TypeVar, cast
 from warnings import showwarning as warn
 
 from proselint import config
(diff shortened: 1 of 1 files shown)


tests/test_config_flag.py
```diff
@@ -29,6 +30,44 @@ def test_deepmerge_dicts() -> None:
...
+
+    assert not any(
+        "typography.punctuation.hyperbole" in path
+        for path in paths
+    )
+
+    assert any(
+        path.startswith("typography.")
+        and not path.startswith("typography.symbols.")
+        and "hyperbole" not in path
```
Nytelife26, line 66:
```suggestion
        and path != "typography.punctuation.hyperbole"
```

tests/test_config_flag.py
```diff
@@ -29,6 +30,44 @@ def test_deepmerge_dicts() -> None:
     }
 
 
+def test_specific_overrides_general() -> None:
```
Nytelife26, line 33: This should technically be part of a `test_registry.py`, which will contain other tests in the future. However, there should be a test for `_flatten_config` here.

proselint/registry/__init__.py
```diff
@@ -54,19 +71,23 @@ def get_all_enabled(
...
-        enabled_checks: list[str] = []
-        skipped_checks: list[str] = []
-        for key, key_enabled in self.enabled_checks.items():
-            (skipped_checks, enabled_checks)[key_enabled].append(key)
+        self.enabled_checks = _flatten_config(enabled)
+        by_specificity = sorted(
+            self.enabled_checks.items(),
+            key=lambda x: x[0].count("."),
+            reverse=True,
+        )
```
Nytelife26, line 79: The configuration being normalised (i.e. flattened) should be a stipulation of this function for performance reasons. It may also be worth requiring the input configuration to be sorted in this way, which can be applied as a post-merge transform like `_flatten_config`.

proselint/registry/__init__.py
```diff
@@ -56,17 +56,15 @@ def get_all_enabled(
         """Filter registered checks by config values based on their keys."""
```
Nytelife26, line 56: The expectation that checks should be sorted by specificity should be documented here.

Nytelife26 (changes requested):

PR #1441, opened 2025-10-02
perf(registry): early exit matches_partial

Nytelife26:
## Relevant issues

Intended to help with #1440.

## Brief

Improve performance of `check.matches_partial` by 85% in the case of a full match using a fast equality, and 66% in all other cases by using `startswith`. 

## Acknowledgements

Credit to @drainpixie for noticing the equivalence of these operations.


Nytelife26:
This is ready for review @drainpixie

PR #1444, opened 2025-10-28
refactor: stabilise wire logging

### LINKED ISSUE #1427: [rfc] stabilised wire output for proselint
## Relevant issues

Follows on from discussions in #1307.

CC: @jclicodes @drainpixie 

## Abstract

Currently, proselint's wire mode (JSON) is unstable, and not guaranteed to produce valid output in the target format. This proposal will resolve this by introducing a standard schema for proselint's wire mode.

Credit to @jclicodes for streamlining the proposed interface.

## Rundown

This section contains a model of proselint's current output in wire mode. You may skip it if you are already familiar with proselint's behaviour.

```ts
interface CheckResult {
    check_path: string;
    message: string;
    start_pos: number;
    end_pos: number;
    replacements?: string;
}

interface LintResult extends CheckResult {
    line: number;
    column: number;
    severity: string;
}

interface ProselintOutput {
    data: { errors: LintResult[] };
    status: "success";
}
```

The `status` key is always `success`, because if proselint errors, the output will not be valid for the specified wire format. This is highly undesirable behaviour for external communications, and [creates errors upstream] in certain situations. `data.errors` is a list of all the warnings generated by proselint over the file. The top-level model `ProselintOutput` is produced once for each file.

[creates errors upstream]: github.com/amperser/proselint/issues/1307

## Proposal

Eventually, it would be great to provide an implementation of the Language Server Protocol. For now, we will set the following goals:

1. Always produce valid wire output, even when proselint encounters an error
2. Produce one model *per-run*, rather than per-file

The wire output will be versioned, and have its version tied to that of proselint itself.

### Infallible operation

To achieve infallibility, proselint will use mutually-exclusive keys with a separate schema for successful and failed operations.

### One output per run

To achieve a single output per-run, rather than per-file, the `data` record will be updated to support multiple keys. Each key will be the name of a lint file, and the success value will be an array of results returned from running
proselint over the file.

### Intended schema

Overall, the new response schema will work as follows, inheriting `LintResult` as previously defined.

```ts
interface ResponseError {
    code: number;
    message: string;
}

interface FileOutput {
    diagnostics?: LintResult[];
    error?: ResponseError;
}

interface ProselintOutput {
    result?: Record<string, FileOutput>;
    error?: ResponseError;
}
```

Both `ProselintOutput` and `FileOutput` use mutually-exclusive keys; that is, `error` is only present if `result` or `diagnostics` respectively are not, and vice versa. One of them must always be present in the output for each type. These two guarantees form a discriminated union, allowing end-users to quickly determine whether proselint failed over an individual file, or altogether. The use of a discriminated union here makes `status` redundant, so it has been removed.

The `result` and `error` objects have been named to correspond with their equivalents in the LSP specification. `error.code` will adhere to the reserved codes in the `ErrorCodes` namespace of the LSP specification, and this proposal will be updated with the proselint-specific codes when possible.

Nytelife26:
## Relevant issues

Resolves #1427.

## Brief

Implement stable infallible JSON logging for Proselint.

## Changes

- Move `OutputFormat` to `proselint.log`
- Use correct output streams for log levels
- Move `LintResult` to `proselint.tools`
- Introduce infallible JSON logging
- Use `span` and `pos` in `CheckResult` and `LintResult`

## Credits

@jclicodes for greatly helping with authoring and refining the wire specification.

### DIFF
--- a/proselint/registry/checks/__init__.py
+++ b/proselint/registry/checks/__init__.py
@@ -98,4 +98,4 @@ class Check(NamedTuple):
         return self.flags.apply(self.check(text), len(text))
 
 
-__all__ = ("Check", "CheckFlags", "CheckResult", "LintResult", "Padding")
+__all__ = ("Check", "CheckFlags", "CheckResult", "Padding")


proselint/log.py
```diff
@@ -3,68 +3,194 @@
...
+                else "%(message)s",
+                "%Y-%m-%d %H:%M:%S",
+            ),
         )
+        self.setLevel(logging.DEBUG if verbose else logging.INFO)
+        self.handlers.clear()
+        self.addHandler(BASE_HANDLER)
+        self.addHandler(err_handler)
+
+    def format_source(self, source: Path | str) -> str:
```
drainpixie: We could use `PathLike[str]`

PR #1454, opened 2025-11-17
Fix crash checking empty string

TomFryersMidsummer:
If `LintFile.__init__` is called with `""` source and content, so `self.content` became `""`, so `_read` assumed `self.source` must be a `Path`.

I've tried to write a test for this, but it doesn't work:
```diff
diff --git a/tests/test_cli.py b/tests/test_cli.py
index 75a09b3..4772150 100644
--- a/tests/test_cli.py
+++ b/tests/test_cli.py
@@ -26,3 +26,7 @@ def test_version(caplog: LogCaptureFixture) -> None:
         assert caplog.record_tuples == [
             ("proselint", logging.INFO, f"Proselint {__version__}")
         ]
+
+
+def test_empty_string_standard_input() -> None:
+    proselint(PARSER.parse_args(("check",)), PARSER)
```

```
OSError: pytest: reading from stdin while output is captured!  Consider using `-s`.
```

However, I've tested it locally, and it does fix the issue.

## Relevant issues

None I could find.

## Brief

Stop Proselint crashing when given an empty file on stdin.

## Changes

Fix crash when checking the empty string on standard input.

### DIFF
--- a/tests/test_cli.py
+++ b/tests/test_cli.py
@@ -1,8 +1,6 @@
 """Verify that the CLI behaves correctly."""
 
 import logging
-import sys
-from io import StringIO
 
 from pytest import LogCaptureFixture, MonkeyPatch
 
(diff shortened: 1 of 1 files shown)


proselint/tools.py
```diff
@@ -160,7 +160,7 @@ def __init__(self, source: str | Path, content: str | None = None) -> None:
 
         self.source = source
 
-        if content:
+        if content is not None:
```
Nytelife26: It strikes me that this is probably the wrong place to make a change. The broken method is `_read`, in its flawed assumption that `self.content` being empty means there must be a source available to read.

Since GitHub doesn't let me propose changes on unchanged regions, I would suggest something like
```py
            if self.content or isinstance(self.source, str):
```
in place of line 192. This should also remove the need for `cast` in `_read`, but I do not have the opportunity to test that currently.

PR #1483, opened 2026-02-25
feat: add speciesist language checks

stuckvgn:
## Summary

Adds a new social_awareness.speciesism check module that identifies speciesist language patterns and suggests clearer, more precise alternatives. This sits alongside the existing social_awareness checks (sexism, lgbtq, nword).

**Two check groups:**

- **check_preferred_forms** -- Flags idioms and metaphors rooted in animal harm and suggests non-speciesist alternatives
- **check_derogatory** -- Flags animal comparisons used as insults or stereotypes

### Examples

| Speciesist phrase | Suggested alternative |
|---|---|
| kill two birds with one stone | solve two problems at once |
| beat a dead horse | belabor the point |
| wild goose chase | futile search |
| let the cat out of the bag | reveal the secret |
| guinea pig | test subject |
| the elephant in the room | the obvious issue |
| dog-eat-dog | cutthroat |
| rat race | daily grind |
| work like a dog | work relentlessly |
| eat like a pig | eat ravenously |
| stubborn as a mule | extremely stubborn |
| blind as a bat | completely blind |
| quiet as a mouse | extremely quiet |

### Why this matters

Speciesist language normalizes the treatment of animals as objects, pests, or inferior beings. Just as proselint already flags sexist and homophobic language, flagging speciesist idioms helps writers choose language that is both more precise and more inclusive.

### Academic references

- Dunayer, J. (2001). Animal Equality: Language and Liberation. Ryce Publishing.
- Stibbe, A. (2001). Language, Power, and the Social Construction of Animals. Society and Animals, 9(2), 145-161. doi:10.1080/10350330109384726
- Bogueva, D., and Marinova, D. (2022). Transforming Language About Animals. In: Handbook of Research on Social Marketing and Its Influence on Animal Origin Food Product Consumption.

### Changes

- proselint/checks/social_awareness/speciesism.py -- New check module with 44 phrase mappings across two check groups
- proselint/checks/social_awareness/__init__.py -- Register the new module
- tests/examples.py -- Add pass/fail test examples

## Test plan

- [x] All existing tests pass (321 passed)
- [x] All structural validation tests pass (231 passed, including batch count, path, and register checks)
- [x] New speciesism examples pass/fail correctly
- [x] Module auto-enabled via existing social_awareness: true config default

PR #1490, opened 2026-04-08
feat: add speciesist language checks

stuckvgn:
Resubmission of #1483 at @Nytelife26's invitation after the original fork was deleted.

## Summary

Adds a `social_awareness.speciesism` check module alongside the existing `sexism`, `lgbtq`, and `nword` checks. Two check groups:

- **check_preferred_forms** — flags idioms rooted in animal harm and suggests clearer alternatives (26 phrases, e.g. "kill two birds with one stone" → "solve two problems at once")
- **check_derogatory** — flags animal comparisons used as insults (9 phrases, e.g. "eat like a pig" → "eat ravenously")

## Changes

- `proselint/checks/social_awareness/speciesism.py` — new check module
- `proselint/checks/social_awareness/__init__.py` — register the new module
- `tests/examples.py` — pass/fail examples

## Why this fits proselint

Just as proselint flags sexist and homophobic language, speciesist idioms are worth flagging: they use animals as stand-ins for negativity or as objects of harm, which is imprecise writing on top of being exclusionary to a growing portion of readers.

## References

- Dunayer, J. (2001). *Animal Equality: Language and Liberation.* Ryce Publishing.
- Stibbe, A. (2001). Language, Power, and the Social Construction of Animals. *Society and Animals*, 9(2), 145–161.
- Bogueva, D., and Marinova, D. (2022). Transforming Language About Animals. In *Handbook of Research on Social Marketing.*

### DIFF
--- a/proselint/checks/social_awareness/speciesism.py
+++ b/proselint/checks/social_awareness/speciesism.py
@@ -10,8 +10,29 @@ date:       2026-04-08 12:00:00
 categories: writing
 ---
 
-Points out speciesist idioms and animal comparisons used as insults,
-and suggests clearer alternatives.
+Speciesism is the assumption of human superiority over other animals,
+expressed linguistically through idioms that treat animals as objects,
+nuisances, or inherently inferior beings. Joan Dunayer's *Animal Equality:
+Language and Liberation* (2001) is a foundational academic text examining
+how everyday language encodes and reinforces this bias.
+
+This check flags two categories of speciesist language:
+
+1. **Speciesist idioms** — common phrases whose imagery frames animals as
+   instruments (e.g. 'kill two birds with one stone'), pests (e.g. 'rat
+   race'), or property (e.g. 'cash cow'). Replacing these with neutral
+   alternatives improves clarity and avoids reinforcing the view that
+   animals exist primarily as means to human ends.
+
+2. **Animal comparisons used as insults** — similes that invoke animals to
+   describe human failings (e.g. 'eat like a pig', 'stubborn as a mule').
+   These expressions demean both the person and the animal, and clearer
+   alternatives are nearly always available.
+
+The idiom list uses Dunayer (2001) as its academic grounding. The specific
+phrases included draw on widely recognised speciesist language patterns in
+English usage; they are not all direct quotations or enumerated examples
+from that text.
 
 """
 
--- a/proselint/checks/social_awareness/speciesism.py
+++ b/proselint/checks/social_awareness/speciesism.py
@@ -1,0 +1,75 @@
+"""
+Speciesism.
+
+---
+layout:     post
+source:     Dunayer, J. (2001). Animal Equality: Language and Liberation.
+source_url: https://doi.org/10.1080/10350330109384726
+title:      speciesism
+date:       2026-04-08 12:00:00
+categories: writing
+---
+
+Points out speciesist idioms and animal comparisons used as insults,
+and suggests clearer alternatives.
+
+"""
+
+from proselint.registry.checks import Check, types
+
+CHECK_PATH = "social_awareness.speciesism"
+
+check_preferred_forms = Check(
+    check_type=types.PreferredFormsSimple(
+        items={
+            "kill two birds with one stone": "solve two problems at once",
+            "beat a dead horse": "belabor the point",
+            "wild goose chase": "futile search",
+            "let the cat out of the bag": "reveal the secret",
+            "guinea pig": "test subject",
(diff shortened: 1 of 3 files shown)


proselint/checks/social_awareness/speciesism.py
```diff
@@ -0,0 +1,75 @@
+"""
+Speciesism.
+
+---
+layout:     post
+source:     Dunayer, J. (2001). Animal Equality: Language and Liberation.
+source_url: https://doi.org/10.1080/10350330109384726
```
Nytelife26: Perhaps it's just me, but the DOI provided here doesn't seem to work. I found an active one for the Cambridge University Press publication at https://doi.org/10.1017/S0962728600024489, or you may use any other archival links.

proselint/checks/social_awareness/speciesism.py
```diff
@@ -0,0 +1,75 @@
...
+title:      speciesism
+date:       2026-04-08 12:00:00
+categories: writing
+---
+
+Points out speciesist idioms and animal comparisons used as insults,
+and suggests clearer alternatives.
```
Nytelife26: Your original PR used the correct citation format and had a much more elaborate explanation section. What was the reason for the change?

proselint/checks/social_awareness/speciesism.py
```diff
@@ -0,0 +1,75 @@
...
+check_preferred_forms = Check(
+    check_type=types.PreferredFormsSimple(
+        items={
+            "kill two birds with one stone": "solve two problems at once",
+            "beat a dead horse": "belabor the point",
+            "wild goose chase": "futile search",
+            "let the cat out of the bag": "reveal the secret",
+            "guinea pig": "test subject",
+            "the elephant in the room": "the obvious issue",
+            "dog-eat-dog": "cutthroat",
```
Nytelife26, line 52: Are these all specific examples from the source itself? I can think of potentially better ways to say many of these, so if so, I'll need to investigate the source further.

proselint/checks/social_awareness/speciesism.py
```diff
@@ -0,0 +1,75 @@
...
+        items={
+            "eat like a pig": "eat ravenously",
+            "stubborn as a mule": "extremely stubborn",
+            "blind as a bat": "completely blind",
+            "quiet as a mouse": "extremely quiet",
+            "drunk as a skunk": "extremely drunk",
+            "fat as a pig": "overweight",
+            "sweating like a pig": "sweating heavily",
+            "slow as a snail": "extremely slow",
+            "memory like a goldfish": "very short memory",
```
Nytelife26: In line with our other checks, we shouldn't recommend "very".

Nytelife26 (changes requested):

Nytelife26:
It occurs to me that we should consider adding the option to specify multiple sources in the headers

cc: @drainpixie

proselint/checks/social_awareness/speciesism.py
Nytelife26: Is there any chance you could address this while I obtain a copy of the source for review, please? Thank you :)

stuckvgn:
Thanks for the detailed feedback @Nytelife26! Pushed an update:

- **DOI**: Already updated to the Cambridge University Press link you provided.
- **Docstring**: Restored a more detailed explanation section in line with other checks in the codebase — describes what speciesist idioms are, why they matter, and what the check does. Also explicitly names the two categories the check covers.
- **Source attribution**: Added a clarifying paragraph at the end of the docstring: the check uses Dunayer (2001) as academic grounding; the idiom list draws on widely recognised speciesist language patterns in English, and the phrases are not all direct quotations or enumerated examples from that text.
- **No 'very'**: Confirmed — no alternatives in the current file use 'very'. (The original first commit had `"memory like a goldfish": "very short memory"`, which was already fixed to `"short memory"` before this review.)

Let me know if anything else needs adjusting.

Nytelife26:
My primary request is related to the check header. You appear to be using APA citation style, rather than MLA 9th Ed., and some of the fields are missing. You may refer to the [verification section of the checks documentation](https://github.com/amperser/proselint/blob/main/docs/checks.md#verification) for further details.

Aside from that, while I agree with the validity of the recommendations here, the checks in Proselint are typically direct recommendations or direct applications of recommendations from prominent authors and writing bodies. If you know of any usage guides or terminology lists that specifically contain the entries used here, it would be good to link to them.

I'm going to spend some time ensuring none of the entries conflict with our other sources and checks in other ways. Following that review, and the other changes I just mentioned, I will be happy to accept this.

